PR #13, opened 2021-06-29
Add the ability to generate streams of scalar type Ion Data

linlin-s:
*Description of changes:*

Guidance for reviewing:
Add class GeneratorOptions.java:
After parsing the arguments from the command line, the generated HashMap will be passed to this class. The relevant data generating method will be executed according to the specifications in the HashMap.

Add class GeneratorOptionsValidator.java:
This class is used for detecting the wrong combinations of the options in the command line.  e.g. The command line ‘ion-java-benchmark generate --data-size 5000 --data-type decimal --timestamps-template “[2021T]” example.10n’ will be detected as an error because the option --timestamps-template is not supported by Ion decimal. 

Add more descriptions to Main.java:
Add usage description of Ion data generator.
Provide the descriptions of the additional options of Ion data generator.
One example of how to use the Ion data generator is provided.

Add class WriteRandomIonValues.java
All methods in this class are used for generating different types of Ion data, all available types are from this set (INT | DECIMAL | STRING | SYMBOL | TIMESTAMP | FLOAT | CLOB | BLOB).


tgregg (approved):
Previously reviewed here: https://github.com/linlin-s/ion-java-benchmark-cli/pull/4

PR #14, opened 2021-07-07
Add the ability to generate Ion Struct based on Ion Schema.

linlin-s:
*Description of changes:*
This PR adds the ability to generate Ion Struct data based on the constraints provided by Ion Schema file, and also fixed some problems from the last PR

Review guide:
GeneratorOptions.java:
Add the logic which can process option --input-ion-shcema. 

GeneratorOptionsValidator.java:
Add constant which contain all options which are invalid when generating Ion Data based on Ion Schema file.

IonSchemaUtilities.java:
This class contains all constants which are relevant to constraint field name in Ion Schema, and methods which can parse the value in these constraints.

ReadGeneralConstraints.java:
This class is used for getting constraints from Ion Schema file and then calling the relevant Ion data generator based on the constraints.

WriteRandomIonValues.java:
Add methods to generate and write Ion Struct to the target file.
Redefine some parameters to more accurate types (instead of Objects) in some existing methods.

textStructs.isl:
This file is used during the test process.


tgregg (approved):
Previously reviewed in https://github.com/linlin-s/ion-java-benchmark-cli/pull/5

PR #15, opened 2021-07-08
Adds the ability to generate Ion List based on the constraints in Ion Schema.

linlin-s:
*Description of changes:*
This PR adds the ability to generate Ion List based on Ion Schema and updates some constraints processing methods to allow them to process more cases.

Review guidance:
pom.xml: 
Added dependency ion-schema-kotlin which will be used in unit test.

GeneratorOptionsValidator.java:
Changed all constant names to snake case.

IonSchemaUtilities.java:
Changed the method processing annotations. Instead of returning annotations as an Ion List, updated method can also return the information of the constraints of annotation(required | optional | ordered).

ReadGeneralConstraints.java:
Added the case to write Ion List 

WriteRandomIonValues.java:
Updated the methods of writing and constructing Ion struct to enable these methods to process more cases when annotations are required, optional or ordered.
Added methods to write and construct Ion List with the constraints in Ion Schema.

DataGeneratorTest.java:
Added unit test for the new feature which can detect whether violation occurs when comparing the generated data with Ion Schema.

testStruct.isl and testList.isl are for test only.


tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -189,13 +198,56 @@ public void testSizeOfGeneratedData() throws Exception {
         assertTrue(difference <= 0.1 * expectedSize);
     }
 
+
+    /**
+     * Test if there's violation by comparing every single data with the Ion schema constraint.
+     * @throws Exception if error occur
+     */
+    @Test
+    public void testViolation() throws Exception {
```
tgregg, line 207: It looks like you'll be able to reuse almost all of this test in order to test other `--input-ion-schema` combinations, right? I see that currently it's only testing structs, but you could move most of this into a separate method and reuse it for lists, etc.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -277,7 +277,7 @@ private static BigDecimal randomSecondWithFraction(Random random, int scale) {
     public static void writeRandomTimestamps(int size, IonType type, String path, String timestampTemplate, String format) throws Exception {
         File file = new File(path);
         try (IonWriter writer = WriteRandomIonValues.formatWriter(format, file)) {
-           WriteRandomIonValues.writeRequestedSizeFile(size, writer, file, type, 0, null, null, null, timestampTemplate, null, null);
+           WriteRandomIonValues.writeRequestedSizeFile(size, writer, file, type, 0, null, null, null, timestampTemplate, null, null, null);
```
tgregg, line 280: Here's a suggestion for how to improve the readability of these method calls that contain tons of nulls. Consider creating helper methods for each type that call through to `WriteRandomIonValues.writeRequestedSizeFile`. The helper method fills in the nulls for the options that aren't relevant for that type.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -726,4 +750,93 @@ public static void constructIonStruct(IonStruct fields, IonList annotations, Ion
...
+     */
+    public static void constructScalarTypeData(IonType valueType, IonWriter writer, int occurTime) throws IOException {
+        for (int i = 0; i < occurTime; i++) {
+            switch (valueType) {
+                // If more scalar types of Ion data are supported, this is the point to add more cases.
+                case STRING:
+                    writer.writeString(WriteRandomIonValues.constructString(defaultRange, 20));
+                    break;
+                case INT:
+                    writer.writeInt(WriteRandomIonValues.constructInt());
```
tgregg, line 838: When you have a method that is known to be incomplete for certain enum values, I always recommend throwing some sort of exception if the method is provided with one of the unsupported enum values. That will make it much easier to figure out what's going on if you try to add support for other scalar types elsewhere but forget to update this method.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -726,4 +750,93 @@ public static void constructIonStruct(IonStruct fields, IonList annotations, Ion
...
+    /**
+     * Construct Ion List based on the constraints in Ion Schema.
+     * @param writer is Ion Writer.
+     * @param constraints is an IonStruct which contains the top-level constraints in Ion Schema.
+     * @throws Exception if errors occur when reading or writing data.
+     */
+    public static void constructIonList(IonWriter writer, IonStruct constraints) throws Exception {
+        // When there's only one required element in Ion List and the length of generated Ion List is not specified, we set the default length as a integer smaller than 20.
+        Random random = new Random();
+        int listSize = random.nextInt(20);
```
tgregg, line 781: At some point we should probably make the max size configurable.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -726,4 +750,93 @@ public static void constructIonStruct(IonStruct fields, IonList annotations, Ion
...
+        int occurTime;
```
tgregg, line 782: "occurrences" might be a better name.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -527,14 +528,42 @@ private static void writeDataToFile(IonType type, IonWriter writer, Random rando
...
             case STRUCT:
-                WriteRandomIonValues.constructIonStruct(fields, annotations, writer);
+                if (annotationMap != null) {
+                    String annotationConstraint = annotationMap.get(IonSchemaUtilities.KEYWORD_CONSTRAINT).toString();
+                    IonList annotationList = (IonList) annotationMap.get(IonSchemaUtilities.KEYWORD_ANNOTATIONS);
+                    if (annotationConstraint.equals(IonSchemaUtilities.KEYWORD_REQUIRED)) {
+                        for (int i = 0; i < annotationList.size(); i++) {
+                            int index = random.nextInt(annotationMap.size());
+                            annotation = annotationList.get(index);
+                            WriteRandomIonValues.constructIonStruct(fields, annotation, writer);
```
tgregg, line 546: Is it acceptable for more than one annotation to be present on a single struct?

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -93,16 +101,25 @@ public static int parseConstraints(IonStruct value, String keyWord) throws IOExc
...
      * @param value is the Ion struct which contain the current constraint field.
-     * @return an Ion List contains all annotations.
+     * @return a HashMap contains the information of annotations, both the constraint of the annotation and the value of the annotation.
      */
-    public static IonList getAnnotation(IonStruct value) {
-        IonList annotationList = null;
-        String annotationConstraint = value.get(keyWordAnnotations).getTypeAnnotations()[0];
-        if (annotationConstraint.equals(keyWordRequired)) {
-            annotationList = (IonList) value.get(keyWordAnnotations);
+    public static Map<String, Object> getAnnotation(IonStruct value) {
```
tgregg, line 108: Should this be `Map<String, IonValue>`?

tgregg (commented):
Cool. Can you paste a couple of examples of the struct and list values generated from the test schemas into a PR comment?

linlin-s:
Examples of Ion Structs based on the Ion Schema file testStruct.isl:
{firstName:"𫗷⍡𢚬𥃯ք낱㦊⢉𠜗𪂚籖𫞐㐗𨍮䭟驿𨚔𣆲軖踦",last_updated:6815T} 
gold_class::{firstName:"𥍪涱衤姼𢾀𩑯썮៏𩎥𨯖𪘴𠖝⒅𧡹쩹雄𪲑䩰𡼋𒂓",last_updated:5874T} 
club_member::{firstName:"𥏽𧡤窪𢁔ꨢ𨄽ܕ齒䯰Ĥュಱ𥭶𤉥𞺹𥾫苼𤶳䆚𩲩",last_updated:8712T}

Examples of Ion Lists based on the Ion Schema file testList.isl:
["澆ꤿ괬䬖ⵥꯩ𨈶𨁭㍶嘅𪀁𝂄忓𦗳䌣𑚞𣤝𫇖𫏃魂",5488134831370622953,368774753466644457,-3231447914290719273,-4569443140516669587,5731707804664757623,8406373166337107299,3495999020345970823]
["㴾𣛁큉𐰟Ď곞滶雿𦹞퍦𢏓𢾺㓋墏鍎츶",-6520895721209821656,-134366090508701836,-1453926387171834629,8387863127643979739]

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -189,13 +198,56 @@ public void testSizeOfGeneratedData() throws Exception {
...
+            Schema newSchema = ISS.newSchema(content);
+            Type type = newSchema.getType(ionSchemaName);
+            while (reader.next() != null) {
+                IonValue value = SYSTEM.newValue(reader);
+                Violations violations = type.validate(value);
+                if (!violations.isValid()) {
+                    System.out.println(value);
+                    System.out.println(violations);
+                }
+                assertTrue(violations.isValid());
```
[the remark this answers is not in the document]
linlin-s: > Instead of using explicit `System.out.println`s, I recommend something like
> 
> ```
> assertTrue("Violations " + violations + "found in value " + value, violations.isValid());
> ```

Change has been updated in the following commit

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -20,14 +26,17 @@
...
+import java.util.stream.Stream;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
 public class DataGeneratorTest {
-
-    public static String outputFile;
+    private static String OUTPUT_FILE;
```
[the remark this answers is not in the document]
linlin-s: > This isn't actually a constant, as it's reassigned in each test.

This line has been changed back.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
linlin-s: > It looks like you'll be able to reuse almost all of this test in order to test other `--input-ion-schema` combinations, right? I see that currently it's only testing structs, but you could move most of this into a separate method and reuse it for lists, etc.

Yes, this testViolation method is valid for other `--input-ion-schema` combinations, and the method which tests the violation when generating Ion List has been added.

linlin-s: > It looks like you'll be able to reuse almost all of this test in order to test other `--input-ion-schema` combinations, right? I see that currently it's only testing structs, but you could move most of this into a separate method and reuse it for lists, etc.

Yes this method is valid for other combinations as well. In order to avoid duplicated codes, a 'for loop' is added in this method, for each loop, the different ion schema file will be tested.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -189,13 +198,56 @@ public void testSizeOfGeneratedData() throws Exception {
...
+    @Test
+    public void testViolation() throws Exception {
+        Map <String, Object> optionsMap = Main.parseArguments("generate", "--data-size", "5000", "--format", "ion_text", "--input-ion-schema", "./tst/com/amazon/ion/benchmark/testStruct.isl", "test8.ion");
+        String inputFilePath = optionsMap.get("--input-ion-schema").toString();
+        Path classSchemaFile = Paths.get(inputFilePath);
+        OUTPUT_FILE = optionsMap.get("<output_file>").toString();
+        try (
+                IonReader readerInput = IonReaderBuilder.standard().build(new BufferedInputStream(new FileInputStream(inputFilePath)));
+                IonReader reader = DataGeneratorTest.executeAndRead(optionsMap);
+                Stream<String> lines = Files.lines(classSchemaFile)
```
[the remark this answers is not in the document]
linlin-s: > reassembled

This replacement has been updated in the following commit.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
linlin-s: > Should this be `Map<String, IonValue>`?

It should be Map<String, Object>, because there one key:value pair which represents the constraint of annotation and the type of constraint is String which cannot be cast to IonValue, and the value of the other key:value pair which represents the real value of annotations is IonList.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > Is it acceptable for more than one annotation to be present on a single struct?

No it is not acceptable, only one annotation should be assigned to a single struct.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > "occurrences" might be a better name.

Changes has been updated in the following commit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > When you have a method that is known to be incomplete for certain enum values, I always recommend throwing some sort of exception if the method is provided with one of the unsupported enum values. That will make it much easier to figure out what's going on if you try to add support for other scalar types elsewhere but forget to update this method.

Added the 'default' case which can throw exceptions when the current enum type is not supported.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > At some point we should probably make the max size configurable.

There's no constraint in Ion List specifies the length of generated list. If configurable max size is supported, one more additional option should be added in the usage.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > Here's a suggestion for how to improve the readability of these method calls that contain tons of nulls. Consider creating helper methods for each type that call through to `WriteRandomIonValues.writeRequestedSizeFile`. The helper method fills in the nulls for the options that aren't relevant for that type.

Inside of the helper method, there will be a process to assign unused parameters to null/0, instead of writing helper methods for each type, how about defining class variables which with default value 'null'? In this way if some specific variables are not used, they will be assigned as null automatically and will avoid writing the same functionalities. Changes will be updated in the following commit, if there's better idea, please feel free to leave comment.

linlin-s (commented):
The next commit contains the updated version of codes and one example about how to generate Ion List based on the constraints of Ion Schema has been added.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -50,6 +51,14 @@
     final static private IonSystem system = IonSystemBuilder.standard().build();
     final static private List<Integer> defaultRange = WriteRandomIonValues.parseRange("[0, 1114111]");
     final static private Timestamp.Precision[] precisions = Timestamp.Precision.values();
```
tgregg, line 53: These should follow the naming style for constants.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
tgregg: Sounds good. The Ion Schema constraint I'm referring to is container_length: https://amzn.github.io/ion-schema/docs/spec.html#container_length

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -50,6 +51,14 @@
...
+    static private int codePointLength = 0;
+    static private List<Integer> pointRange;
+    static private List<Integer> expValRange;
+    static private List<Integer> coefficientDigitRange;
+    static private String timestampTemplate;
+    static private IonStruct fields;
+    static private Map<String, Object> annotationMap;
+    static private IonStruct constraintStruct;
```
tgregg, line 61: I like the idea here to improve readability. Since you never assign these, I'd make two small changes:
1. Make them `final` and explicitly assign them to `null`.
2. Name them like `NO_CODE_POINT_LENGTH` or `NO_FIELDS` (using the naming style for constants).

This makes it clear that they're intended to be used to improve readability and don't need to be assigned.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -204,39 +206,34 @@ public void testSizeOfGeneratedData() throws Exception {
...
-            IonDatagram schema = ReadGeneralConstraints.LOADER.load(readerInput);
-            String ionSchemaName = null;
-            for (int i = 0; i < schema.size(); i++) {
-                IonValue schemaValue = schema.get(i);
-                if (schemaValue.getType().equals(IonType.STRUCT) && schemaValue.getTypeAnnotations()[0].equals(IonSchemaUtilities.KEYWORD_TYPE)) {
-                    IonStruct constraintStruct = (IonStruct) schemaValue;
-                    ionSchemaName = constraintStruct.get(IonSchemaUtilities.KEYWORD_NAME).toString();
-                    break;
+    public void testViolationOfIonStruct() throws Exception {
+        for (int index = 0; index < SCHEMA_FILES.size(); index++) {
```
tgregg, line 210: This is great to see that you're reusing this code for all of your test schemas. The only problem with this technique is that if one of the files starts failing, it's not immediately clear which one because all of them are tested in the same test method. Therefore, I recommend moving all of the code inside the loop into a reusable method that accepts the schema file, and creating a separate `@Test` method for each file, which will simply call into the method. That way it's always clear which file(s) caused test failures and it's easy to target just the tests for the failing files for debugging.

Another option is [parameterized testing](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests), which is great way to achieve this using JUnit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
tgregg: This is better. I've added a small suggestion related to this on the new revision.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
tgregg: Do you mean that Ion Schema doesn't allow users to specify that a struct may have multiple annotations?

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
tgregg: Good, I added a comment related to this on the new revision.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > Sounds good. The Ion Schema constraint I'm referring to is container_length: https://amzn.github.io/ion-schema/docs/spec.html#container_length

Probably instead of adding an option in the usage, adding one method to process this general constraint would be a better way to make the max size configurable, which can also be applied when generating other types of Ion Data,  thanks for the idea.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > Do you mean that Ion Schema doesn't allow users to specify that a struct may have multiple annotations?

I review the specification of Ion Schema found that there's no way to specify whether the value has one or multiple annotations in the schema annotations constraint, but both ways are acceptable by default(assign the generated value with only one or multiple annotations). However in the method above, only one annotation was assigned to each value. And I think the reorganization of logic of this part is necessary. Changes will not be included in the following commit but will be updated in the next PR.

linlin-s (commented):
For the minor updates, changes will be included in the following commit. For adding the ability to specify the container_length and rewrite the logic of assigning annotation parts, changes will be included in the next PR.

tgregg (approved):
Nice!

PR #17, opened 2021-07-24
Adds the ability to compare benchmark results from different ion-java commits.

linlin-s:
*Description of changes:*
This PR adds the ability of comparing benchmark results from different ion-java commits and adds the ability to generate Ion Struct in a nested format. There's a unit test added for this comparison API.

**Motivation:**

There's a GitHub actions workflow which will be triggered when a new commit is pushed to the master branch of Ion java to help developers detect the performance regression. One way to detect the performance regression is to compare the benchmark result from the new commit with the one from the existing commit. The comparison process in the workflow will be finished by this API. After comparing process, the workflow will keep working on regression detecting process.

**Review guide:**
- Main.java:
In this class, a usage 'ion-java-benchmark compare (--benchmark-result-previous <file_path>) (--benchmark-result-new <file_path>) <output_file>' was added, and description for this usage was added in the following section.
- ParseAndCompareBenchmarkResults.java:
This is the new class which can process the comparison process. When the ion-java-benchmark-cli invoke contains keyword 'compare', the methods in this class will be called. After calculating the relative changes for each aspect, the final result will be written in an Ion Struct format.
- WriteRandomIonValues.java
Add one more case for method constructIonStruct which provide the ability to generate nested Ion struct.
- DataGeneratorTest.java 
Add unit test for comparing benchmark results API. This test method will read the output file generated by the API and compare the result from the output file with the accurate results.
Add unit test for generating nested Ion struct.
- testNestedStruct.isl
This file contains Ion Schema which specifies the constraints of Ion Data in a nested structure. 
- IonLoaderBenchmarkResultNew.ion
This file is the benchmark result from the new ion-java commit. This file be used in the unit test.
- IonLoaderBenchmarkResultPrevious.ion
This file is the benchmark result from the existing ion-java commit. This file be used in the unit test.


linlin-s:
The result from benchmark results comparison process:
{input:"testIonData.10n",params:{format:"ION_BINARY",type:"FILE",api:"DOM"},relative_difference_score:{'·gc.alloc.rate':-0.002851051607559,'Heap usage':0.0219662830324952,'Serialized size':0.000000,speed:0.017441311589927}}

Examples of generated random nested Ion Struct:
corporate::{firstName:"𪍝া釖ᒰ蒨𝃠鏴𝝍잴🅎⠕쉱곃⩢𦈣ﲯ⾻𥢛羏𠫹",last_updated:3428T,addresses:["𪁵ᦴ阍ﴃ𨷔𩡖𧓴씯쎥G皎讕𢉫煈滑𠶀ܧ𡝭","𢴫𣲣𥘘𧒞殼笅톺騑𒋻𥆡ᢚ𥾗쥨𖧋쥈𫆃𦴘𩅼쐂魧","𦈇𠝠𢛞⩴𧿩䈆阷糊𩊻𐰑籑ﮩ𪕤𧇨挄𧾎刽𦠹𢦂𢴝","𪑍𖠭欳墡𪇁ᥨ𠛺𡋩嵠朣ⴈ给♂飧𢻤𨁐䮓𦄦𦭗𢼵","Ｋ竤匎俏𖥺མ嚖𝟙𤡒𢺂𣗁𡲾京𦖃⒩𨰇","㘑𩹆ⷃ譤獘착𣓽𦢱ꋩ鴥椾𤼼ⰺ𫈄𠇆墆𦘥𪡋𢛰䱇","𦩁𢙙𪉤䜤䋑𩍖䖾𑚮𥝧퉪랺𧰲윖𠝯⑃噽𥴳徇𫘿ᴫ"]} 
corporate::club_member::{firstName:"🜆ൄ㼖䝧𫍪𠀗ϵ瓔𩞈𢖆𝍂뭀巉𦴅𧮴𝙂혖𥱛㳜𠹳",last_updated:1774T,addresses:["𤑝𨚾눃䃶𦲠棥㯰𥡦𩙿嚧𒇙栄欣𣥁𡌽𦛝︕𤸉黌㟜","徹𪇆𦞑𣚋赒𦄚ᴏ㡄ꉺ𡶝㈵⊟𨰗𨷂쐝嫞𢾍乷碌𣸸","𪙢ꃎ黆꾣𥮦뺲𠹓恂𡀉䢕れ𢯖洜𦮰𡑱ꉝ𤅪𤨤从𧧀","ⶮ𠡿𥩋𒆈𒐱ꘂ撇䘸⎩䄝즮蠚𒅫𤧚놶삏ꔾ코繹","暅𝟤巡𤻼햊矟ᕞ⦰篶ᆰ𠤀孥ꨍ༜꺤♬䭯㧻ₛ𠘼"]}

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -0,0 +1,172 @@
...
+     * This is a helper method which create a IonReader for benchmark result and extract the IonStruct which contain parameters or scores information.
+     * @param benchmarkResultFilePath is the path of benchmark result.
+     * @return an Ion Struct which contains information of parameters using during the benchmark process or scores.
+     * @throws Exception if error occurs when reading Ion Data.
+     */
+    private static IonStruct readHelper(String benchmarkResultFilePath) throws Exception {
+        try (IonReader reader = IonReaderBuilder.standard().build(new BufferedInputStream(new FileInputStream(benchmarkResultFilePath)))) {
+            reader.next();
+            reader.stepIn();
+            reader.next();
```
tgregg, line 109: There should be some checking on the `next()` calls to make sure the reader is positioned on the expected type.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Write calculated relative changes of scores in an Ion Struct into the generated file.
+     * @param benchmarkResult is the path of benchmark result.
+     * @param outputFilePath is destination path of generated result.
+     * @param scoreMap is a hashmap which match score with the aspect it represents.
+     * @throws Exception if error occurs when reading Ion data.
+     */
+    private static void writeResult(String benchmarkResult, String outputFilePath, Map<String, BigDecimal> scoreMap) throws Exception {
+        File file = new File(outputFilePath);
+        String inputFileName = getParameter(benchmarkResult, INPUT).toString();
```
tgregg, line 135: I believe if you cast this to `IonString` and use `stringValue()` that will give you the raw string value, so you won't need to use `inputFileName.replace("\"", "")` later.

tgregg (commented):
Nice! Just two minor comments.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > I believe if you cast this to `IonString` and use `stringValue()` that will give you the raw string value, so you won't need to use `inputFileName.replace("\"", "")` later.

That worked, thanks for suggestion, changes will be updated in the next commit

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > There should be some checking on the `next()` calls to make sure the reader is positioned on the expected type.

Sounds good. Changes will be updated in the next commit.

linlin-s (commented):
The next commit fixed all problems mentioned in the first commit.

tgregg (approved):

PR #18, opened 2021-07-29
Adds the abilities to organize the generated benchmark results and detect the performance regression of ion-java.

linlin-s:
*Description of changes:*

**Motivation:**

In the GitHub workflow which can detect the performance regression of ion-java from the new commit, there is one step which requires generating benchmark results from different ion-java commits by using three different shapes of test Ion data as input and six different ion-java-benchmark-cli invokes for each test data. In order to executes all expected options combinations of ion-java-benchmark and make the workflow organized and easier to understand, this step will be finished by this additional java API (GenerateAndOrganizeBenchmarkResults.java)

**Review Guide:**

**1. GenerateAndOrganizeBenchmarkResults.java :**  
In GitHub actions workflow, this API will be called twice for different versions of ion-java. The benchmark result from the first invoke will be named as 'previous.ion', and the benchmark result from the second time will be named as 'new.ion'. 

**generateAndSaveBenchMarkResults:**
- First of all, this method will iterate all test data files in the given directory. For each test data, there will be six ion-java-benchmark-cli invokes. The generated benchmark results generating by invoking the same ion-java-benchmark command will be saved under the same directory which is named as the aggregated version of ion-java-benchmark invoke. Before writing benchmark result into the specific directory, there is one step to check if the expected file is already exist, if exist the benchmark result will be named as 'new.ion' which means this method has been called before and generated a benchmark result called 'previous.ion'.  The workflow will not save the generated file if we don't upload it into artifact, so we don't need to worry about if there is a directory with the same name exist or not.

**2. Main.java:**
- Added more usages and the descriptions of the additional options.

**3. ParseAndCompareBenchmarkResults.java:**
- Added method detectRegression which can detect whether regression happened. This method will compare the report from comparison process with the threshold, if the relative change of the score in the report is smaller than provided threshold(all threshold provided are negative which represents the relative decrease score), then the performance regression message will be printed, else "Regression not detected" will be printed. The threshold.ion file contains all threshold scores from different ion-java-benchmark invokes, all data are valid Ion structs which have the same structure as the Ion data in comparison report. After getting the comparison report, this method will go through threshold.ion to find the data which has the same parameters as the Ion struct in comparison report, then finishing the regression detecting process.

**4. Added files testList.10n, testStruct.10n and testNestedStruct.10n are test Ion data which will be used in method GenerateAndOrganizeBenchmarkResults**


src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -175,5 +179,48 @@ private static void writeResult(String benchmarkResult, String outputFilePath, M
...
+                // Get the fields which contain parameters information.
+                thresholdScoresStruct.remove(RELATIVE_DIFFERENCE_THRESHOLD);
+                // Get the fields which contain threshold scores information.
+                IonStruct thresholds = (IonStruct) thresholdsStruct.get(RELATIVE_DIFFERENCE_THRESHOLD);
+                if (thresholdScoresStruct.equals(comparisonResultStruct)) {
+                    for (String keyWord : scoreMap.keySet()) {
+                        IonDecimal thresholdScore = (IonDecimal) thresholds.get(keyWord);
+                        if (scoreMap.get(keyWord).compareTo(thresholdScore.bigDecimalValue()) < 0) {
+                            String regressionDetect = "The performance regression detected when benchmark the ion-java from the new commit with the test data: " + comparisonResultStruct.get(INPUT) + " and parameters: " + comparisonResultStruct.get(PARAMETERS);
+                            return regressionDetect;
```
tgregg, line 217: You can just return the string literal; no need to assign to a variable and immediately return it. (Same comment about the other return from this method.)

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -175,5 +179,48 @@ private static void writeResult(String benchmarkResult, String outputFilePath, M
...
+    /**
+     * Compare the relative changes of benchmark results with the thresholds, if the relative change smaller than threshold score which represent the decrease threshold of one
+     * specific aspect, then the performance regression detected and return 'true'.
+     * @param thresholdFilePath specifies the path of file which contains all threshold scores.
+     * @param scoreMap is a hashmap which match relative change of the score with the aspect it represents.
+     * @param outputFilePath is the destination of generated report after comparison process.
+     * @return a String which contains the information about whether performance regression happened.
+     * @throws Exception if occur happen when reading Ion Data.
+     */
+    private static String detectRegression(String thresholdFilePath, Map<String, BigDecimal> scoreMap, String outputFilePath) throws Exception {
```
tgregg, line 196: I'd like to see a unit test for this method to make sure it detects regressions when expected.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -175,5 +179,48 @@ private static void writeResult(String benchmarkResult, String outputFilePath, M
...
+                        }
+                    }
+                }
+            }
+        }
+        String regressionNotDetected = "Regression not detected";
+        return regressionNotDetected;
     }
-}
+}
```
tgregg, line 226: Add a newline at the end of the file.

tst/com/amazon/ion/benchmark/threshold.ion
```diff
@@ -0,0 +1,19 @@
+{input:"testStruct.10n",params:read::{format:"ION_BINARY",type:"FILE",api:"DOM"},relative_difference_score_threshold:{'·gc.alloc.rate':0.0000,'Heap usage':-0.01614003182541480555801658106555348482343250635890,'Serialized size':0.000000,speed:-0.0913183279742764692953173302096577494641544817982732}}
```
tgregg, line 1: So a negative score means that `new` is _slower_ than `previous`, right? Therefore, you check to see if the actual score is more negative than the threshold.

I recommend reducing the precision of the thresholds. There are very few measurements in the universe that require 50 digits of precision. Let's round to 4 digits.

src/com/amazon/ion/benchmark/GenerateAndOrganizeBenchmarkResults.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class GenerateAndOrganizeBenchmarkResults {
+
+    final public static List<String> BENCHMARK_OPTION_COMBINATIONS = Arrays.asList(
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar read --api dom --results-format ion ",
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar read --api streaming --results-format ion ",
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar write --api streaming --results-format ion ",
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar read --api streaming --format json --results-format ion ",
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar read --api dom --format json --results-format ion ",
+        "java -jar target/ion-java-benchmark-cli-0.0.1-SNAPSHOT-jar-with-dependencies.jar write --api streaming --format json --results-format ion ");
```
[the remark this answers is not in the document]
linlin-s: > It would be great if we could retrieve the version (in this case `0.0.1`) from a centralized location so we don't have to change this code in the event that the CLI version changes. This may be complicated to do since the version number is declared in pom.xml, so at the very least we should store the version string in a constant so that if it does need to be changed in code it can be done on one line. We may also want to store the entire jar invocation in a constant (`java -jar target/...jar`) so that it can be reused here and below in `generateAndSaveBenchMarkResults`.

This can be achieved by parsing pom.xml as long as the position of version information in pom.xml will not be changed. Pom parser will be added in the following commit.

linlin-s (commented):
This commit fixed all problems from the last one.

src/com/amazon/ion/benchmark/GenerateAndOrganizeBenchmarkResults.java
```diff
@@ -62,10 +82,28 @@ public static void generateAndSaveBenchMarkResults(Map<String, Object> optionsMa
...
         }
     }
+
+    /**
+     * Parse pom.xml and extract the version of ion-java-benchmark-cli.
+     * @param filePath the file path of pom.xml/
+     * @return the version of ion-java-benchmark-cli in a string format.
+     * @throws Exception if errors occur when parsing data.
+     */
+    private static String pomParser(String filePath) throws Exception {
```
tgregg, line 100: This method should be named something like "parseVersionFromPom".

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -213,14 +216,18 @@ private static String detectRegression(String thresholdFilePath, Map<String, Big
...
                         }
                     }
                 }
             }
+            if (regressions.size() != 0) {
+                return "The performance regression detected when benchmark the ion-java from the new commit with the test data: "
+                        + fileName.substring(fileName.lastIndexOf(File.separator) + 1) + " and parameters: " + comparisonResultStruct.get(PARAMETERS) + System.lineSeparator()
+                        + "The following aspects have regressions: " + regressions + System.lineSeparator();
+            } else {
+                return System.lineSeparator();
```
tgregg: Is returning the empty string `""` sufficient?

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -294,6 +306,40 @@ public void testParseBenchmark() throws Exception {
         }
     }
 
+    /**
+     * Test whether the detecting regression process can return the expected result.
+     * In this unit test we use an Ion file which contain regression on [·gc.alloc.rate] as input to test the detectRegression method.
+     * @throws Exception if error occur when reading Ion data.
+     */
+    @Test
+    public void testRegressionDetecting() throws Exception {
```
tgregg, line 315: (Apologies if this comment appears twice; it looks like it was lost when I tried to post it)

This verifies that a regression is detected when it is present, but we should also test that a regression is _not_ detected when it is _not_ present. You'll be able to reuse most of the code from this test.

src/com/amazon/ion/benchmark/GenerateAndOrganizeBenchmarkResults.java
linlin-s: > This method should be named something like "parseVersionFromPom".

Sounds good. Change will be updated.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > sufficient

Return the empty string "" can make sure that only regression detection messages will be printed in the workflow which will be easier for users to tract the regression detected message. Because there is extra empty line in the workflow, 'if condition' will be added in the following commit to avoid printing empty line.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
linlin-s: > (Apologies if this comment appears twice; it looks like it was lost when I tried to post it)
> 
> This verifies that a regression is detected when it is present, but we should also test that a regression is _not_ detected when it is _not_ present. You'll be able to reuse most of the code from this test.

No problem. Additional unit test will be added.

linlin-s (commented):
The following commit fixed all problems from the second commit.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -335,20 +336,54 @@ public void testRegressionDetecting() throws Exception {
...
+        for (String keyWord : ParseAndCompareBenchmarkResults.BENCHMARK_SCORE_KEYWORDS) {
+            IonValue score = scoresStruct.get(keyWord);
+            if (score.getType().equals(IonType.FLOAT)) {
+                IonFloat scoreFloat = (IonFloat) score;
+                scoreMap.put(keyWord, scoreFloat.bigDecimalValue());
+            } else {
+                IonDecimal scoreDecimal = (IonDecimal) score;
+                scoreMap.put(keyWord, scoreDecimal.bigDecimalValue());
+            }
+        }
```
tgregg, line 371: All of this can be shared with `testRegressionDetected` by refactoring it into a method, right?

tgregg (approved):

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
linlin-s: > All of this can be shared with `testRegressionDetected` by refactoring it into a method, right?

Yes, the duplicated part has been refactored in the following commit.

tgregg (approved):

PR #21, opened 2021-08-06
Adds a method to calculate thresholds instead of using hard-code thresholds.

linlin-s:
**Descriptions of changes:**

**Motivations:**
- Fixed the bug in Ion Data Generator which blocked generating most scalar types of Ion Data.
- Change the way of getting threshold scores for performance detection process in order to reduce the false negative. Instead of providing an Ion List which contains hard-code thresholds, the new methods added are able to calculate threshold for each ion-java-benchmark-cli invoke.

**Review guide:**

- **GeneratorOptionsValidator.java:**
  - Fixed the bug that cause problems when generating scalar types of Ion Data.

- **Main.java:**

  - Delete the option '--threshold' in the usage 'compare'. In this case the comparison process don not need to take threshold as input, it will calculate threshold score for each ion-java-benchmark invoke instead.

- **ParseAndCompareBenchmarkResults.java:**

  - Added methods to calculate and construct threshold map from benchmark results. 
    - We will calculate threshold scores from benchmark results of ion-java from existing commit and new commit. In the method getThresholdMap, we compare the thresholds and get the smaller value as the final threshold. (We need to do some experiments on getting threshold process, so far this method reduced the false negative and provide the expected results).

- **DataGeneratorTest.java:**

  - Added unit test for the method getThresholdMap to make sure this method will resturn the result as we expected.

- **testComparisonReport.ion:**

  - Edited one aspect of benchmark result to make performance regression in order to test whether the detectRegression method will return the results as expected.


src/com/amazon/ion/benchmark/GeneratorOptionsValidator.java
```diff
@@ -53,6 +53,12 @@ public static void checkValid(String [] args, Map <String, Object> optionsMap) {
                 case DECIMAL:
                     GeneratorOptionsValidator.throwException(INVALID_FOR_DECIMAL, commandLine);
                     break;
+                case INT:
+                case FLOAT:
+                case BLOB:
+                case CLOB:
+                case SYMBOL:
+                    break;
```
tgregg, line 61: Does this mean we silently do nothing for these types?

tgregg (approved):

src/com/amazon/ion/benchmark/GeneratorOptionsValidator.java
linlin-s: > Does this mean we silently do nothing for these types?

Thanks for catching this! For these types there should be one step to check if the invalid options provided. I'll add it in the next commit.

tgregg (approved):

PR #31, opened 2022-04-27
Adds features to process constraint 'valid_values' and generate float and symbol from ion schema file.

linlin-s:
**Description of changes:**

_IonSchemaUtilities.java_
- parseConstraints:
   Changed the method of getting data from constraints when `min` or `max` provided.
- checkValidRange: 
   This method is used for checking whether the constraint `valid_value` has annotation `range`. If there is annotation  `range`, this method will return an IonList which contains the range of `valid_value`. This method will be used as an additional specification in the process of constructing data.
- parseValidValue:
  This method is used for parsing the constraint `valid_value` which only provided a list of IonValue without the annotation range. This method will be used in the first step of writing data. If provided ISL struct contains this constraint, the write-data-to-file process will not call other constructing data methods.

_WriteRandomIonValues.java_
- writeDataToFile:
  Added a step to check if a list of valid value has been provided in the constraint struct. If there is provided valid value, the writer will write the randomly selected data from the `valid_value` list to the file.
- constrcutFloat:
  This method is used for constructing double value. If the constraint `valid_value: range::[<number>]` provided, this method will construct value based on the specified valid range, else this method will construct double value randomly.
- constrcutInt:
  Added steps to process the constraint `valid_value: range::[<number>]`.

**Test:**

_DataGeneratorTest.java_
- Added test for generating float.
- Added test for generating float with specified `valid_value` range.
- Added test for generating symbol.
- Added test for generating string.
- Added test for generating int with specified `valid_value` range.

 

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +168,77 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
     }
 
+    /**
+     * This method is used for checking whether the constraint 'valid_value' has the annotation 'range'.
+     * If 'valid_value' has the annotation 'range', this method will return the range value in IonList format, else it will return null.
+     * @param constraintStruct is the Ion struct which contain the current constraint field.
+     * @return the result of checking whether constraint 'valid_value' contains the annotation 'range'.
+     * @throws Exception if error occurs when building the IonReader.
+     */
+    public static IonList checkValidRange(IonStruct constraintStruct) throws Exception {
```
desaikd, line 178: The method name here doesn't seem to identify with what it is doing. Maybe a name like `getValidValuesAsRange()` would work.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -45,6 +46,7 @@ public class IonSchemaUtilities {
     public static final String KEYWORD_MAX = "max";
     public static final String KEYWORD_SCALE = "scale";
     public static final String KEYWORD_PRECISION = "precision";
+    public static final String KEYWORD_VALID_VALUE = "valid_value";
```
desaikd, line 49: As per Ion schema specification `valid_values` is correct the constraint name. 
Reference : https://amzn.github.io/ion-schema/docs/spec.html#valid_values

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +168,77 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
+        IonSystem SYSTEM = IonSystemBuilder.standard().build();
+        IonLoader LOADER = SYSTEM.newLoader();
+        if (constraintStruct != null) {
+            try (IonReader reader = IonReaderBuilder.standard().build(constraintStruct)) {
+                reader.next();
+                reader.stepIn();
+                while (reader.next() != null) {
+                    if (reader.getFieldName().equals(KEYWORD_VALID_VALUE) && reader.getTypeAnnotations().length != 0) {
+                        // If the constraint contains the annotation 'range' it will return the range value in IonList format for the further data construction process.
+                        IonDatagram datagram = LOADER.load(reader);
```
desaikd, line 189: Can we add a check for `range` here? This doesn't seem to check the annotation `range::`. There could be invalid annotations `range::`, `abc::` etc as Ion schema allows for open content.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +168,77 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    /**
+     * This method is used for paring constraint 'valid_value: [ <VALUE>... ]' and return a randomly selected IonValue.
+     * This method only process the constraint 'valid_value' without the annotation 'range'.
```
desaikd, line 200: Does this mean it only parses for non range `valid_values`?  `without annotation range` makes it difficult to understand teh statement. Maybe add a comment as below would help:
```suggestion
     * This method only parses for `valid_values` that doesn't contain ranges.
```

tst/com/amazon/ion/benchmark/testFloatValidValue.isl
```diff
@@ -0,0 +1,9 @@
+schema_header::{
+}
+type::{
+    name: Float,
+    type: float,
+    valid_value: range:: [-1.447, 2.457],
```
desaikd, line 6: I am unsure how this test works. But Are we testing the generated data by validating it against the schema?
If yes, how does this test pass? Ion schema would not be able to understand `valid_value` constraint as it names as `valid_values`.

desaikd (commented):
This is coming along well! There are some concerning changes I have addressed below. 
Mainly, are we testing the the generated data against the schema? Because `valid_value` is not there in the Ion schema grammar but `valid_values` is. Hence, seems like these tests would pass by ignoring the `valid_value` constraint as open content.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
linlin-s: > The method name here doesn't seem to identify with what it is doing. Maybe a name like `getValidValuesAsRange()` would work.

Sounds good.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
linlin-s: > Can we add a check for `range` here? This doesn't seem to check the annotation `range::`. There could be invalid annotations `range::`, `abc::` etc as Ion schema allows for open content.

Yes, I assumed that the `range::` is the only possibility. I'll add it in the following commit.

tst/com/amazon/ion/benchmark/testFloatValidValue.isl
linlin-s: > I am unsure how this test works. But Are we testing the generated data by validating it against the schema? Ion schema would not be able to understand `valid_value` constraint as it names as `valid_values` and would consider it open content.

Yes, the first step of testing is to load the ion schema file and then test the generated data by validating it against the schema. If the grammar isn't right it suppose to throw errors. I'll double check the test method and make edition in the follow commit.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +168,77 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
+                        result = (IonList)datagram.get(0);
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    /**
+     * This method is used for paring constraint 'valid_value: [ <VALUE>... ]' and return a randomly selected IonValue.
```
tgregg, line 200: paring -> parsing

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +168,77 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
+     * This method only process the constraint 'valid_value' without the annotation 'range'.
+     * @param constraintStruct is the Ion struct which contain the current constraint field.
+     * @return an IonValue selected from the provided list randomly.
+     * @throws Exception if error occurs when building the IonReader.
+     */
+    public static IonValue parseValidValue(IonStruct constraintStruct) throws Exception {
+        IonSystem SYSTEM = IonSystemBuilder.standard().build();
```
tgregg, line 207: We can reuse one IonSystem and one IonLoader across all methods in the library.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -299,37 +301,48 @@ public static void writeRequestedSizeFile(int size, IonWriter writer, File file,
...
-                writer.writeClob(WriteRandomIonValues.constructLobs(constraintStruct));
-                break;
-            default:
-                throw new IllegalStateException(type + " is not supported.");
+        // Check whether the valid_value constraints provided in the top level constraint struct. If a list of valid_value provided, the generated
+        // value should be selected randomly from the provided valid value list every iteration.
+        // Constraint 'valid_value' has two formats, with annotation '' and without annotation ''. This step only check the format that without annotation.
+        // If the annotation 'range' has been detected, it will be processed in the constructing data steps.
+        if (constraintStruct != null && IonSchemaUtilities.parseValidValue(constraintStruct) != null && IonSchemaUtilities.checkValidRange(constraintStruct) == null) {
+            IonValue value = IonSchemaUtilities.parseValidValue(constraintStruct);
```
tgregg: Consider restructuring this to avoid having to call `parseValidValue` twice.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -341,13 +354,14 @@ private static void writeDataToFile(IonWriter writer, IonStruct constraintStruct
...
         String constructedString;
+        Integer codePointsLengthBound;
         String regexPattern = IonSchemaUtilities.parseTextConstraints(constraintStruct, IonSchemaUtilities.KEYWORD_REGEX);
-        Integer codePointsLengthBound = IonSchemaUtilities.parseConstraints(constraintStruct, IonSchemaUtilities.KEYWORD_CODE_POINT_LENGTH);
         if (regexPattern != null) {
             RgxGen rgxGen = new RgxGen(regexPattern);
             constructedString = rgxGen.generate();
         } else {
             Random random = new Random();
+            codePointsLengthBound = IonSchemaUtilities.parseConstraints(constraintStruct, IonSchemaUtilities.KEYWORD_CODE_POINT_LENGTH);
```
tgregg, line 364: What if there's both a regex and a code point length constraint?

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -368,9 +382,32 @@ public static String constructString(IonStruct constraintStruct) throws Exceptio
...
+     * Construct the float which is conformed with the constraints provided in ISL.
+     * @param constraintStruct is an IonStruct which contains the top-level constraints in Ion Schema.
+     * @return the constructed double value.
+     * @throws Exception
+     */
+    public static double constructFloat(IonStruct constraintStruct) throws Exception {
+        Random random = new Random();
+        double randomDouble;
+        if (IonSchemaUtilities.checkValidRange(constraintStruct) != null) {
+            IonList range = IonSchemaUtilities.checkValidRange(constraintStruct);
```
tgregg: Consider restructuring this to avoid having to call `checkValidRange` twice.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -395,11 +432,24 @@ public static BigDecimal constructDecimal(IonStruct constraintStruct) throws Exc
...
+        RandomDataGenerator randomDataGenerator = new RandomDataGenerator();
         Random random = new Random();
-        long longValue = random.nextLong();
+        long longValue;
+        if (IonSchemaUtilities.checkValidRange(constraintStruct) != null) {
+            IonList range = IonSchemaUtilities.checkValidRange(constraintStruct);
+            // Convert IonValue to long
+            long lowerBound = Long.valueOf(range.get(0).toString());
+            long upperBound = Long.valueOf(range.get(1).toString());
+            longValue = randomDataGenerator.nextLong(lowerBound, upperBound);
```
tgregg, line 449: We can achieve this behavior fairly easily without taking a dependency on `org.apache.commons.math3.random.RandomDataGenerator`. I recommend doing that unless there's another good reason for this dependency.

linlin-s: > We can achieve this behavior fairly easily without taking a dependency on `org.apache.commons.math3.random.RandomDataGenerator`. I recommend doing that unless there's another good reason for this dependency.

Applied this suggestion in the following commit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > What if there's both a regex and a code point length constraint?

In the following commit, the logic of constructing string will be categorized into three conditions:
1. ISL contains both `regex` and `codepoint_length`:
   First, generating string based on the provided `regex` then check whether the generated string aligned with the constraint `codepoint_length`. If not, the method will keep generating until the generated string matches both constraints. It might take a while to generate the string which matches two constraints at the same time or the provided constraints are conflict with each other (e.g. regex: "^B[0-9]{9}$", codepoint_length: range:: [1, 3]). Under this consideration, preset the times of implementing generating string step at most 10. When the iterations hit 10 times but not find any matches string, it will write null value.
2. Only contain one of the constraints:
   Construct the string conformed with the provided constraint.
3. No constraint:
   Under this condition, the method will construct string randomly with random Unicode codepoint and `codepoint_length`.

Note:
1. The upper bound of iteration time requires further discussion.
2. The unit test cannot be implemented until this issue be closed

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > Consider restructuring this to avoid having to call `parseValidValue` twice.

Updated in the following commit.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
linlin-s: > We can reuse one IonSystem and one IonLoader across all methods in the library.

Updated in the following commit.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +171,72 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
         return constraint;
     }
 
+    /**
+     * This method is used for getting the constraint value which contains annotation 'range'.
+     * @param constraintStruct is the Ion struct which contain the current constraint field.
+     * @return the result of checking whether constraint 'valid_value' contains the annotation 'range'.
```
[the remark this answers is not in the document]
linlin-s: Yes, this method could be used by other constraints which contains range too. All methods with the functionalities of understanding ion schema will be refactored in the next PR and refactoring this method is part of the plan. There will be a class QuantifiableConstraint to handle all constraints which contain range.

src/com/amazon/ion/benchmark/IonSchemaUtilities.java
```diff
@@ -166,6 +171,72 @@ public static String parseTextConstraints(IonStruct constraintStruct, String key
...
+    public static IonValue parseValidValue(IonStruct constraintStruct) throws Exception {
+        Random random = new Random();
+        IonValue result = null;
+        try (IonReader reader = IonReaderBuilder.standard().build(constraintStruct)) {
+            reader.next();
+            reader.stepIn();
+            // Build the top level constraints into a reader and iterate the struct to find the constraint 'valid_value'.
+            while (reader.next() != null) {
+                // Find the constraint 'valid_value' and get the value of this constraint.
+                // The formats of 'valid_value' are [ <VALUE>... ], <RANGE<NUMBER>>, <RANGE<TIMESTAMP>>.
```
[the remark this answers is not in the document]
linlin-s: Yes this method only parse `valid_values: [<VALUE>... ] `. I'll update all typo of `valid_values` in the following commit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -299,37 +302,51 @@ public static void writeRequestedSizeFile(int size, IonWriter writer, File file,
...
-                break;
-            case CLOB:
-                writer.writeClob(WriteRandomIonValues.constructLobs(constraintStruct));
-                break;
-            default:
-                throw new IllegalStateException(type + " is not supported.");
+        IonValue value = null;
+        // Check whether the valid_value constraints provided in the top level constraint struct.
+        // If a list of valid_value provided, the generated value should be selected randomly from the provided valid value list every iteration.
+        // Constraint 'valid_value' has two formats, with annotation '' and without annotation ''. This step only check the format that without annotation.
```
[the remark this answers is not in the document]
linlin-s: The statement will be updated in the next commit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -299,37 +302,51 @@ public static void writeRequestedSizeFile(int size, IonWriter writer, File file,
...
+        // If the annotation 'range' has been detected, it will be processed in the constructing data steps.
+        if (constraintStruct != null) {
+            value = IonSchemaUtilities.parseValidValue(constraintStruct);
+        }
+        if (value != null && IonSchemaUtilities.getConstraintValueAsRange(constraintStruct, IonSchemaUtilities.KEYWORD_VALID_VALUE) == null) {
```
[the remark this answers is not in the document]
linlin-s: > 

Thanks for suggestions. This is also part of the consideration of refactoring. The refactoring will be contained in the next PR.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -395,11 +457,23 @@ public static BigDecimal constructDecimal(IonStruct constraintStruct) throws Exc
...
+    public static long constructInt(IonStruct constraintStruct) throws Exception {
         Random random = new Random();
-        long longValue = random.nextLong();
+        long longValue;
+        IonList range = IonSchemaUtilities.getConstraintValueAsRange(constraintStruct, IonSchemaUtilities.KEYWORD_VALID_VALUES);
+        if (range != null) {
+            // Convert IonValue to long
+            long lowerBound = Long.valueOf(range.get(0).toString());
+            long upperBound = Long.valueOf(range.get(1).toString());
+            longValue = lowerBound + (long)(Math.random() * (upperBound - lowerBound));
```
tgregg, line 473: You already have a `Random` instance, and should probably use that here instead of `Math.random()` (which creates its own `Random` instance). That should allow you to use `nextLong()` to avoid the cast.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -211,6 +216,51 @@ public void testViolationOfIonBlob() throws Exception {
...
+     * Test if there's violation when generating floats from Ion Schema file.
+     * @throws Exception if error occurs during the violation detecting process.
+     */
+    @Test
+    public void testViolationOfFloat() throws Exception {
+        DataGeneratorTest.violationDetect(INPUT_ION_FLOAT_FILE_PATH);
+    }
+
+    /**
+     * Test if there's violation when generating floats from Ion Schema file.
```
tgregg, line 229: floats -> symbols

This same copy/paste error occurs on other JavaDoc comments on the test methods in this file.

tgregg (approved):
Two minor comments. Approved pending resolution of those and approval by @desaikd.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
linlin-s: > floats -> symbols
> 
> This same copy/paste error occurs on other JavaDoc comments on the test methods in this file.

Thanks, the updates will be included in the next commit.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
linlin-s: > You already have a `Random` instance, and should probably use that here instead of `Math.random()` (which creates its own `Random` instance). That should allow you to use `nextLong()` to avoid the cast.

To avoid this situation, the new updates in the following commit use `ThreadLocalRandom.current().nextLong(lowerBound, upperBound) ` to generate the random value which allows to set range.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -340,37 +356,82 @@ private static void writeDataToFile(IonWriter writer, IonStruct constraintStruct
...
+        int codePoint;
+        do {
+            codePoint = random.nextInt(DEFAULT_RANGE.get(1) - DEFAULT_RANGE.get(0) + 1) + DEFAULT_RANGE.get(0);
+            type = Character.getType(codePoint);
+        } while (type == Character.PRIVATE_USE || type == Character.SURROGATE || type == Character.UNASSIGNED);
+        return codePoint;
+    }
+
+    /**
+     *  Construct string which is conformed with the provided codepoint_length.
```
desaikd, line 396:
```suggestion
     * Construct string which is conformed with the provided codepoint_length.
```

desaikd (approved):
Looks good! Some minor comments.

src/com/amazon/ion/benchmark/WriteRandomIonValues.java
```diff
@@ -340,37 +356,82 @@ private static void writeDataToFile(IonWriter writer, IonStruct constraintStruct
      * @throws Exception if error occurs when parsing the constraints.
      */
     public static String constructString(IonStruct constraintStruct) throws Exception {
+        Random random = new Random();
         String constructedString;
         String regexPattern = IonSchemaUtilities.parseTextConstraints(constraintStruct, IonSchemaUtilities.KEYWORD_REGEX);
         Integer codePointsLengthBound = IonSchemaUtilities.parseConstraints(constraintStruct, IonSchemaUtilities.KEYWORD_CODE_POINT_LENGTH);
-        if (regexPattern != null) {
+        // For now, if there are potentially-conflicting constraints detected, an exception statement will be thrown.
```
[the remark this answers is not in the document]
linlin-s: > 

Sounds good, I'll open an issue and add the link here.

PR #37, opened 2022-07-11
Adds features to process constraint 'element' and generate IonSexp.

linlin-s:
*Issue #, if available:*

N/A

***Description of changes:***
This PR added features to generate `IonSexp` and also contains the feature to process constraint `element`. We test the validation of the generated data by providing test ISL files which contain the relevant constraints.

**DataConstructor.java:**

- constructSequenceTypeData:
Added the logic of generating `IonSexp`. `IonSexp` generation process shared the same logic with `IonList` generation.
In this method, the first step is to check if there are potentially conflict constraints in the provided ISL file and then generating data based on the provided constraints. The newly added logic is processing constraint `element`. The way of processing `element` is to parse the value of `element` into `ReparsedType` then pass the re-parsed value to method `constructIonData` to generate data recursively. Constraint `element` can also work with constraint `container_length` (if it is contained in the ISL file) together, and constraint `container_length` helps decide the length of generated container type data.
- constructIonStruct:
In this method we also added logic to process the constraint `element`. If `element` provided in `IonStruct` ISL file, then the field values in the generated `IonStruct` would be aligned with constraint `element`, and the filed name would be a random string with default `codepoint_length` range `[0, 20)`.

**ReparsedType.java:**

- Added case to parse the constraint `element`.

**Element.java:**

- This is the newly added class which is used for parsing constraint `element`.

**Test:**

- Added test case of generating `IonStruct` which is conformed with the ISL contains constraint `element`.
- Added test case to check whether there is violation while generating `IonSexp`.

### DIFF
--- a/src/com/amazon/ion/benchmark/DataConstructor.java
+++ b/src/com/amazon/ion/benchmark/DataConstructor.java
@@ -17,6 +17,8 @@ package com.amazon.ion.benchmark;
 
 import com.amazon.ion.IonList;
 import com.amazon.ion.IonReader;
+import com.amazon.ion.IonSequence;
+import com.amazon.ion.IonSexp;
 import com.amazon.ion.IonStruct;
 import com.amazon.ion.IonSystem;
 import com.amazon.ion.IonTimestamp;
@@ -26,6 +28,7 @@ import com.amazon.ion.IonWriter;
 import com.amazon.ion.Timestamp;
 import com.amazon.ion.benchmark.schema.ReparsedType;
 import com.amazon.ion.benchmark.schema.constraints.Contains;
+import com.amazon.ion.benchmark.schema.constraints.Element;
 import com.amazon.ion.benchmark.schema.constraints.Fields;
 import com.amazon.ion.benchmark.schema.constraints.OrderedElements;
 import com.amazon.ion.benchmark.schema.constraints.QuantifiableConstraints;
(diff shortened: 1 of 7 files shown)


src/com/amazon/ion/benchmark/DataConstructor.java
```diff
@@ -246,60 +255,76 @@ public static IonValue constructIonData(ReparsedType parsedTypeDefinition) {
...
-     * Constructing IonList which is aligned with the constraints provided in the constraintMap.
+     * Constructing IonSequence value which is aligned with the constraints provided in the constraintMap.
      * @param constraintMapClone collects the constraints from ISL file, the key represents the name of constraints,
      * and the value is constraint value in ReparsedConstraint format.
-     * @return the constructed IonList value.
+     * @param container represents one of the element from set (IonList | IonSexp).
+     * @return the constructed IonSequence value.
      */
-    private static IonList constructIonList(Map<String, ReparsedConstraint> constraintMapClone) {
+    private static IonValue constructSequenceTypeData(Map<String, ReparsedConstraint> constraintMapClone, IonSequence container) {
```
tgregg, line 297: Should this return `IonSequence`? Or, should this be a templated method that returns the same type as `container`?

tst/com/amazon/ion/benchmark/testSchemaContainsElement.isl
```diff
@@ -0,0 +1,6 @@
+type::{
+    name: StructWithElement,
+    type: struct,
+    element: { type: timestamp, timestamp_precision: range::[year, day]},
+    container_length: 4,
```
tgregg, line 5: Do we have a file that covers the case where `element` is specified but `container_length` is not?

tgregg (approved):

desaikd (approved):
LGTM!

tst/com/amazon/ion/benchmark/testSchemaContainsElement.isl
linlin-s: > Do we have a file that covers the case where `element` is specified but `container_length` is not?

The ISL file which specifies `element` without specifying `container_length` will be added in the following commit.

src/com/amazon/ion/benchmark/DataConstructor.java
linlin-s: This method should return IonSequence for now.  It will be updated in the following commit.

PR #38, opened 2022-07-14
Adds feature to process the constraint 'annotations'.

linlin-s:
*Issue #, if available:*

***Description of changes:***
This PR adds class `Annotations` to process the constraint `annotations`. After parsing the constraint value to `Annotations`, we are able to get a list of annotations or `null` based on the annotation `required | optional | ordered `of the constraint.

**DataConstructor.java:**

- In the method `constructIonData`, instead of returning the value directly for each type case, overwriting the result variable which stores the constructed Ion data. And then check whether the constraint `annotations` is provided in the same level of data structure. If the constraint `annotations` provided, the `annotations` value will be added to the constructed data and be returned.

**ReparsedType.java**

- Adds the case  `ANNOTATIONS` to the method of parsing constraint value to `ReparsedConstraint`.

**Annotations:**

- This class is used for parsing the value of constraint `annotations`. This class provides an attribute that allows users to get the processed `annotations` value in an IonList format.

**Test:**
Adds a test case that cover the constraint `annotations` .

### DIFF
--- a/src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
+++ b/src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
@@ -1,0 +1,74 @@
+package com.amazon.ion.benchmark.schema.constraints;
+
+import com.amazon.ion.IonList;
+import com.amazon.ion.IonValue;
+
+import java.util.Arrays;
+import java.util.Random;
+
+/**
+ * This class aims to process the constraint 'annotations'. After parsing the constraint value to Annotations object, we
+ * are able to get the processed annotations in an IonList format.
+ */
+public class Annotations implements ReparsedConstraint {
+    private IonList annotations;
+
+    /**
+     * Initializing the newly created Annotations object.
+     * @param field represents the value of constraint 'annotations'.
+     */
+    private Annotations(IonValue field) {
+        this.annotations = parseAnnotations(field);
+    }
+
+    /**
+     * Helping access the private attribute 'annotations'.
+     * @return the attribute 'annotations'.
+     */
+    public IonList getAnnotations() {
+        return this.annotations;
+    }
+
+    /**
+     * Process the annotation of the constraint 'annotations'. By default, individual annotations
+     * are optional and this default may be overridden by annotating the annotations list with 'required'.
+     * If annotations must be applied to value in specified order, the list may be annotated with 'ordered'.
+     * This method will process these annotations and return the value aligned with these specifications.
+     * @param field represents the value of constraint 'annotations'.
+     * @return the list of annotations that are used for annotating IonValue.
+     */
+    private IonList parseAnnotations(IonValue field) {
+        String[] annotationsSpecificationList = field.getTypeAnnotations();
+        // If the constraint 'annotation' is not annotated or annotated with "optional",
+        if (annotationsSpecificationList.length == 0 || Arrays.asList(annotationsSpecificationList).contains("optional")) {
+            return randomlyReturnAnnotations(field);
+        } else {
+            return (IonList)field;
+        }
+    }
+
+    /**
+     * Process the constraint 'annotations' without annotation or contains 'optional' annotation.
+     * @param field represents the value of constraint 'annotations'.
+     * @return a null value or a list of annotations randomly.
+     */
+    private IonList randomlyReturnAnnotations(IonValue field) {
+        Random random = new Random();
+        int value = random.nextInt(2);
+        switch (value) {
+            case 1:
+                return (IonList)field;
+            default:
+                return null;
+        }
+    }
+
+    /**
+     * Parsing the value of constraint 'annotations' into Annotations.
+     * @param field represents the value of constraint 'annotations'.
+     * @return the parsed object Annotations.
+     */
+    public static Annotations of(IonValue field) {
+        return new Annotations(field);
+    }
+}
--- a/src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
+++ b/src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
@@ -39,8 +39,13 @@ public class Annotations implements ReparsedConstraint {
      */
     private IonList parseAnnotations(IonValue field) {
         String[] annotationsSpecificationList = field.getTypeAnnotations();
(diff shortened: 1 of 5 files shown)


jobarr-amzn (approved):

src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param field represents the value of constraint 'annotations'.
+     * @return the list of annotations that are used for annotating IonValue.
+     */
+    private IonList parseAnnotations(IonValue field) {
+        String[] annotationsSpecificationList = field.getTypeAnnotations();
+        // If the constraint 'annotation' is not annotated or annotated with "optional",
+        if (annotationsSpecificationList.length == 0 || Arrays.asList(annotationsSpecificationList).contains("optional")) {
+            return randomlyReturnAnnotations(field);
+        } else {
+            return (IonList)field;
```
desaikd, line 46: I don't think annotations would make much contribution to the shape of generated data but I have mentioned some supported features for `annotations` as per [Ion schema spec](https://amzn.github.io/ion-schema/docs/spec.html#annotations) in the `Features` section in this comment.

**_If your intention here is to not consider/ignore these features of `annotations` for the sake of generating data which I think is fine because it might not impact much on the shape of data, then we should atleast keep a note of this in a comment here specifying these features are not considered._**

Features:
`annotations` constraint can have annotation for each listed annotation (`required` or `optional`).
Also, `annotations` list can also have annotation (`required` ,`closed` and `ordered`). 
e.g.
```
annotations: [red, required::green, blue]
annotations: required::[red, optional::green, blue]
annotations: required::ordered::[one, optional::two, three]
annotations: closed::required::[red, green, blue] // Annotations must contain exactly "red", "green", and "blue" in any order
annotations: closed::ordered::[red, green, blue] // Annotations must contain exactly "red", "green", and "blue" in that order
annotations: closed::[red, blue] // Only the annotations "red" and "blue" are permitted, but they are not required
annotations: closed::[] // No annotations are permitted
```

src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
```diff
@@ -0,0 +1,74 @@
...
+     * are optional and this default may be overridden by annotating the annotations list with 'required'.
+     * If annotations must be applied to value in specified order, the list may be annotated with 'ordered'.
+     * This method will process these annotations and return the value aligned with these specifications.
+     * @param field represents the value of constraint 'annotations'.
+     * @return the list of annotations that are used for annotating IonValue.
+     */
+    private IonList parseAnnotations(IonValue field) {
+        String[] annotationsSpecificationList = field.getTypeAnnotations();
+        // If the constraint 'annotation' is not annotated or annotated with "optional",
+        if (annotationsSpecificationList.length == 0 || Arrays.asList(annotationsSpecificationList).contains("optional")) {
```
desaikd, line 43: `optional` at list-level for annotations is not really supported by [Ion schema specification](https://amzn.github.io/ion-schema/docs/spec.html#annotations). By default annotations are optional.
e.g.
```
annotations: [red, green, blue]  // this is by default optional at list level 
```

src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
```diff
@@ -0,0 +1,74 @@
...
+     * Process the annotation of the constraint 'annotations'. By default, individual annotations
+     * are optional and this default may be overridden by annotating the annotations list with 'required'.
+     * If annotations must be applied to value in specified order, the list may be annotated with 'ordered'.
+     * This method will process these annotations and return the value aligned with these specifications.
+     * @param field represents the value of constraint 'annotations'.
+     * @return the list of annotations that are used for annotating IonValue.
+     */
+    private IonList parseAnnotations(IonValue field) {
+        String[] annotationsSpecificationList = field.getTypeAnnotations();
+        // If the constraint 'annotation' is not annotated or annotated with "optional",
```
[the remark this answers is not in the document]
desaikd: Also, you might need to remove the `optional` from here as it is not supported at list-level for `annotations`.
```suggestion
        // If the constraint 'annotations' is not annotated
```

src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
linlin-s: > I don't think annotations would make much contribution to the shape of generated data but I have mentioned some supported features for `annotations` as per [Ion schema spec](https://amzn.github.io/ion-schema/docs/spec.html#annotations) in the `Features` section in this comment.
> 
> **_If your intention here is to not consider/ignore these features of `annotations` for the sake of generating data which I think is fine because it might not impact much on the shape of data, then we should atleast keep a note of this in a comment here specifying these features are not considered._**
> 
> Features: `annotations` constraint can have annotation for each listed annotation (`required` or `optional`). Also, `annotations` list can also have annotation (`required` ,`closed` and `ordered`). e.g.
> 
> ```
> annotations: [red, required::green, blue]
> annotations: required::[red, optional::green, blue]
> annotations: required::ordered::[one, optional::two, three]
> annotations: closed::required::[red, green, blue] // Annotations must contain exactly "red", "green", and "blue" in any order
> annotations: closed::ordered::[red, green, blue] // Annotations must contain exactly "red", "green", and "blue" in that order
> annotations: closed::[red, blue] // Only the annotations "red" and "blue" are permitted, but they are not required
> annotations: closed::[] // No annotations are permitted
> ```

Thanks for the explanation. These features do not have significant impact on the generated ion data when the use case is performance testing. For now, I'll put a comment on that as the future development reference.

src/com/amazon/ion/benchmark/schema/constraints/Annotations.java
linlin-s: > `optional` at list-level for annotations is not really supported by [Ion schema specification](https://amzn.github.io/ion-schema/docs/spec.html#annotations). By default annotations are optional. e.g.
> 
> ```
> annotations: [red, green, blue]  // this is by default optional at list level 
> ```

Thanks for correction. This will be updated in the following commit.

desaikd (approved):

Issue #39, opened 2022-07-14
Adds internal default constraints in Ion Data Generator.


linlin-s:
In order to generate data which is more similar to the real-world data, users are required to provide highly-specified schema. It would be better to provide a set of default constraints internally which are able to be overridden by the users.
i.e.
Input type definition:
```
 type::{
       name: NestedList,
       type: list,
        ordered_elements: [
            string,
            int,
            {type: int, occurs: range::[0, 10] },
            {type: struct,
                fields: {
                first_name: string,
                last_name: string,
                last_updated: { type: timestamp, timestamp_precision: year, occurs: range:: [2, 5]},
                }
            },
        ]
  }
```
Generated data:
```
[
  "𢋎썾䛰𡁜𦓱𓂙緻𧕊赀𤀿딟𨡠𤄧𣀜𝝔鸬𢙕ࠟ",  1257713305652284209,  -5644910875031391494,  -7360934801865493629,  2015045502948171567,  -3426578700639114343,  4799305136895312052,  {
    last_updated: 7238T,    last_updated: 4437T,    last_updated: 5424T,    last_updated: 9091T,    last_name: "𠯽𨓐𩑩熌𥙨峾鴊𠎖𪡤𦜤ה𠝹ቕ",    first_name: "浃␄輬썂𣧱븅"  }
]
```
From the example, the generated data are not widely used in real world. In that case, it might be better to have default constraints to specify the generated data and make the data more practical.

PR #41, opened 2022-07-26
Updates default constraints of data generators.

linlin-s:
*Issue #, if available:*
[ Adds internal default constraints in Ion Data Generator.](https://github.com/amzn/ion-java-benchmark-cli/issues/39)

*Description of changes:*

In order to generate data which is able to mimic the real-world data better, this PR updated default constraints of `String`, `Symbol`, `Timestamp` and `Int`.

**DataConstructor.java:**

- getCodePoint:
According to ASCII, in this method we added logic to narrow the `unicode` code point of generated character to the ranges `[65, 90]` and `[97, 122]` which matches to the character `[A-Z]` and `[a-z]`.
- constructInt:
Added logic to generate integers which are mostly conformed with the range `[0, 1024]`.
- constructTimestamp:
Updated the default precision of generated timestamp as `Day`.

**Test:**

- Added unit test for IonInt generating when there is no constraint provided in ISL.

### DIFF
--- a/src/com/amazon/ion/benchmark/DataConstructor.java
+++ b/src/com/amazon/ion/benchmark/DataConstructor.java
@@ -70,9 +70,12 @@ class DataConstructor {
     // The constant defined below are used as placeholder in the method WriteRandomIonValues.writeRequestedSizeFile.
     final static private IonSystem SYSTEM = IonSystemBuilder.standard().build();
     final static private List<Integer> DEFAULT_RANGE = Arrays.asList(0, 0x10FFFF);
-    final static public Timestamp.Precision[] PRECISIONS = Timestamp.Precision.values();
     final static public IonStruct NO_CONSTRAINT_STRUCT = null;
     final static private int DEFAULT_PRECISION = 20;
+    // The ASCII_CODE_LOWERCASE_A represents the ASCII code of character "a".
+    final static private int ASCII_CODE_LOWERCASE_A = 97;
+    // The ASCII_CODE_UPPERCASE_A represents the ASCII code of character "A".
+    final static private int ASCII_CODE_UPPERCASE_A = 65;
     final static private int DEFAULT_SCALE_LOWER_BOUND = -20;
     final static private int DEFAULT_SCALE_UPPER_BOUND = 20;
     final static private int DEFAULT_CONTAINER_LENGTH = 20;
@@ -396,12 +399,12 @@ class DataConstructor {
     private static int getCodePoint() {
         int index = ThreadLocalRandom.current().nextInt(20);
         int randomIndex = ThreadLocalRandom.current().nextInt(26);
-        if (index >= 0 && index < 10) {
+        if (index < 10) {
             // Randomly generate the unicode of character from [A-Z].
-            return randomIndex + 65;
+            return randomIndex + ASCII_CODE_UPPERCASE_A;
         } else {
             // Randomly generate the unicode of character from [a-z].
-            return randomIndex + 97;
+            return randomIndex + ASCII_CODE_LOWERCASE_A;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/com/amazon/ion/benchmark/DataConstructor.java
```diff
@@ -390,18 +390,19 @@ public static String constructString(Map<String, ReparsedConstraint> constraintM
...
-        int type;
-        int codePoint;
-        do {
-            codePoint = random.nextInt(DEFAULT_RANGE.get(1) - DEFAULT_RANGE.get(0) + 1) + DEFAULT_RANGE.get(0);
-            type = Character.getType(codePoint);
-        } while (type == Character.PRIVATE_USE || type == Character.SURROGATE || type == Character.UNASSIGNED);
-        return codePoint;
+        int index = ThreadLocalRandom.current().nextInt(20);
+        int randomIndex = ThreadLocalRandom.current().nextInt(26);
+        if (index >= 0 && index < 10) {
```
tgregg, line 399: Is `index >= 0` always true due to the contract of `nextInt()`?

src/com/amazon/ion/benchmark/DataConstructor.java
```diff
@@ -390,18 +390,19 @@ public static String constructString(Map<String, ReparsedConstraint> constraintM
...
+            // Randomly generate the unicode of character from [A-Z].
+            return randomIndex + 65;
```
tgregg, line 401: Please store the magic numbers `65` and `97` as constants with comments to describe what they are.

src/com/amazon/ion/benchmark/DataConstructor.java
```diff
@@ -512,8 +521,8 @@ public static Timestamp constructTimestamp(Map<String, ReparsedConstraint> const
         Range range = DEFAULT_TIMESTAMP_IN_MILLIS_DECIMAL_RANGE;
         // Preset the local offset.
         Integer localOffset = localOffset(random);
-        // Preset the default precision.
-        Timestamp.Precision precision = PRECISIONS[random.nextInt(PRECISIONS.length)];
+        // Preset the default precision as 'Day'.
+        Timestamp.Precision precision = PRECISIONS[2];
```
tgregg, line 525: `= Timestamp.Precision.DAY;` would be simpler.

tgregg (approved):

tgregg (approved):

tst/com/amazon/ion/benchmark/testIntWithoutConstraint.isl
```diff
@@ -0,0 +1,4 @@
+type::{
+    name: Int,
+    type: int,
+}
```
desaikd, line 4: Should we also add tests for String, Symbol, Timestamp without any constraints except `type` constraint?
e.g.
```
type::{
    name: my_string,
    type: string,
} 
```

desaikd (approved):

tst/com/amazon/ion/benchmark/testIntWithoutConstraint.isl
linlin-s: > Should we also add tests for `String`, `Symbol`, `Timestamp` without any constraints except `type` constraint? e.g.
> 
> ```
> type::{
>     name: my_string,
>     type: string,
> } 
> ```

The reason we added test for `int` without any constraints except `type` constraint is that there is no specified constraints to narrow the `int` value. For the rest of data types, the functionalities of constraint `type` and other constraints do not work with each other(the data validation process has already covered these tests), so the additional tests for these data types might be duplicated with the exist unit tests.

src/com/amazon/ion/benchmark/DataConstructor.java
linlin-s: > Is `index >= 0` always true due to the contract of `nextInt()`?

Yes, this condition will be deleted in the next commit.

src/com/amazon/ion/benchmark/DataConstructor.java
linlin-s: > `= Timestamp.Precision.DAY;` would be simpler.

This will be updated in the next PR.

PR #42, opened 2022-08-01
Update comparison methods and refactor test data files.

linlin-s:
*Issue #, if available:*
N/A

***Description of changes:***

This PR refactors the test data files of Ion Data Generator and ion-java-regression-detection workflow and save them under different folders. In order to improve the result visualization of the workflow, this PR updates the return value of the performance comparison methods.

**ParseAndCompareBenchmarkResults.java:**

- writeResult:
This method is used for writing the performance comparison result to file. In order to display the regression result to the `job summary `in GitHub workflow, we updated the output value of this method.
- detectRegression:
This method is used for comparing the regression score with threshold score to determine if regression occurs. Instead of returning the regression detection message, we return the detected regression score map directly. This score map will be displayed in the `job summary` session of GitHub workflow.  

**Test:**

- Update the unit test and make the logic aligned with the condition that when the return value of method `detectRegression` is a regression score map. 
- Add three ISL files under the folder workflow which are used for generating test data in the workflow.

**Example:**

No regression detected:

- Before updating the comparison methods:
After running ion-java-performance-regression-detection workflow, if there is no regression detected, when we select the `summary` session, no information will be displayed. 
Screen Shot 2022-08-02 at 8 23 08 AM

-  After updating the comparison methods:
When the workflow receive the output of the method is "no regression detected" for every benchmark process. Selecting the `summary` session, then "No regression detected!" will be displayed.
Screen Shot 2022-08-02 at 8 31 40 AM

Regression detected:

- Before updating the comparison methods:
After running ion-java-performance-regression-detection workflow, if there the regression detected, users need to go back to the workflow to check which benchmark process has performance regression.
Screen Shot 2022-08-02 at 8 39 23 AM

regression-before
-  After updating the comparison methods:
We are able to get the regression information directly from the job `summary` session. This form represents which benchmark process has regression and its relative regression scores.
Screen Shot 2022-08-02 at 9 10 00 AM

_Note: The example screenshots provided above are trying to provide the visualization updates. The link of the workflow execution example will be added to this comment after merging this PR._


tgregg (commented):
Can you please attach before/after screenshots of the regression detection workflow results? That would make it much easier to visualize this change.

linlin-s:
> Can you please attach before/after screenshots of the regression detection workflow results? That would make it much easier to visualize this change.

Yes, I'll add an example of the workflow.

tgregg (approved):
Nice

PR #45, opened 2022-08-22
Adds features to process constraints 'any_of' and 'one_of'.

linlin-s:
*Issue #, if available:*
N/A

*Description of changes:*
This PR adds logic to process the constraints `any_of` and `one_of` and adds the relative unit tests for the new features.

**ReadGeneralConstraints.java:**
Constructing `ReparsedType` object every time we call the method of constructing `IonData`. In this case, if there is constraint `any_of` or `one_of` in the schema, the attributes (`constraintMap`) of  `ReparsedType` will be reconstructed/updated and be aligned with the content of the current type reference.

**DataConstructor.java:**
Changing the strategy of getting` type_name` of the current `type_reference`. This change is aligned with the update of `ReparsedType.java`.

**ReparsedType.java:**

- Adds logic to parse the value of constraint ‘type’ to TypeName object then construct key value pair of `<type, ReparsedConstraint>` then add it to the `constraintMap` which will be passed to the data constructing process. The reason of replacing the original static method `getType` to the current strategy is to make sure the type name passed into the data constructing process is accurate and up-to-date. After adding the logic of processing constraints `any_of`/ `one_of` (the value of these fields is list of the `type_reference`), each `type_reference` will be parsed as `ReparsedType`. Adding the `TypeName` key value pair as one of the constraint in `constraintMap` will make the current `TypeName` aligned with the value of constraint `type` in the type reference.

- handleField:
Adds logic to process the constraints `one_of` and `any_of`. According to Ion Schema Specification, the definition of `[any_of](https://amzn.github.io/ion-schema/docs/spec.html#any_of)` is different from `[one_of](https://amzn.github.io/ion-schema/docs/spec.html#one_of)` while validating ion data. However, in term of generating ion data from schema, these two constraints have the same functionality and would share the same logic. The value of constraint `any_of`/`one_of` is a list of `type_reference`, and one of the `type_reference` will be randomly chosen and then will be parsed to `ReparsedType`. The constraints contained in the randomly picked `type_reference` will be added to the top level `constraintMap` which will be used in the following data constructing process.

- Adds case of processing constraint `content`. When using Dataguide to infer schema from real-world data, the type definition of struct contains `content: closed`. For this constraint, there is no additional logic needed for the following data constructing process.

**TypeName.java:**
This class is used for parsing the value of constraint `type` as `TypeName`, and the attribute of this object represents the type name in `IonType` format.

**Test:**
Adds tests for the feature of processing constraints `any_of` and `one_of`.

### DIFF
--- a/src/com/amazon/ion/benchmark/schema/constraints/TypeName.java
+++ b/src/com/amazon/ion/benchmark/schema/constraints/TypeName.java
@@ -1,0 +1,33 @@
+package com.amazon.ion.benchmark.schema.constraints;
+
+import com.amazon.ion.IonType;
+import com.amazon.ion.IonValue;
+
+public class TypeName implements ReparsedConstraint {
+    private IonType typeName;
+
+    /**
+     * Initializing the newly created TypeName object.
+     * @param field represents the value of constraint 'type'.
+     */
+    private TypeName(IonValue field) {
+        this.typeName = IonType.valueOf(field.toString().toUpperCase());
+    }
+
+    /**
+     * Helping access the private attribute 'type'.
+     * @return the attribute 'type'.
+     */
+    public IonType getTypeName() {
+        return this.typeName;
+    }
+
+    /**
+     * Parsing the value of constraint 'type' into TypeName.
+     * @param field represents the value of constraint 'type'.
+     * @return the parsed object TypeName.
+     */
+    public static TypeName of(IonValue field) {
+        return new TypeName(field);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/com/amazon/ion/benchmark/schema/constraints/TypeName.java
```diff
@@ -0,0 +1,33 @@
+package com.amazon.ion.benchmark.schema.constraints;
+
+import com.amazon.ion.IonType;
+import com.amazon.ion.IonValue;
+
+public class TypeName implements ReparsedConstraint {
```
tgregg, line 6: Please add a class-level JavaDoc to describe what a TypeName is.

tgregg (approved):
Looks good to me!

PR #51, opened 2022-11-02
Adds option to allow generate the same random data.

linlin-s:
*Issue #, if available:*
N/A

*Description of changes:*
This PR adds an option `--seed` which allows users to specify the seed value when generating random ion data. If the provided seed value and the schema file are the same as the last execution, the same randomized ion data will be generated.

**GeneratorOptions.java**
Add a static field `randomSeed` to the class `GeneratorOptions`. This field will be overwrote in the step of processing the `--seed` option. If the seed value provided, the random instance will be crated with the specified seed value and be used in the following data generation process. Otherwise, the random instance will be created with a random seed value. 

**Main.java**
Add an option `--seed` which allow users to specify the seed value. When they re-run the command with the same seed value and the same input ion schema, the same randomized ion data will be generated.

**Test**
- Add unit test to test whether the generated randomized data is as same as the last execution when specifying the seed value.
- Add unit test to test if the regenerated data is randomized when there is no seed value provided. 

### DIFF
--- a/src/com/amazon/ion/benchmark/GeneratorOptions.java
+++ b/src/com/amazon/ion/benchmark/GeneratorOptions.java
@@ -10,7 +10,7 @@ import java.util.Random;
  * Execute Ion Data Generator after receiving the hashmap of command line options.
  */
 public class GeneratorOptions {
-    public static Random randomSeed;
+    public static Random random;
 
     /**
      * Check the validation of input ion schema and execute the Ion Data generating process.
@@ -23,9 +23,9 @@ public class GeneratorOptions {
         String path = optionsMap.get("<output_file>").toString();
         String inputFilePath = optionsMap.get("--input-ion-schema").toString();
         if (optionsMap.get("--seed") != null) {
-            randomSeed = new Random(Long.valueOf(optionsMap.get("--seed").toString()));
+            random = new Random(Long.valueOf(optionsMap.get("--seed").toString()));
         } else {
-            randomSeed = new Random();
+            random = new Random();
         }
         // Check whether the input schema file is valid and get the loaded schema.
         Schema schema = IonSchemaUtilities.loadSchemaDefinition(inputFilePath);
--- a/src/com/amazon/ion/benchmark/GeneratorOptions.java
+++ b/src/com/amazon/ion/benchmark/GeneratorOptions.java
@@ -4,11 +4,13 @@ import com.amazon.ionschema.Schema;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 
 /**
  * Execute Ion Data Generator after receiving the hashmap of command line options.
  */
 public class GeneratorOptions {
+    public static Random randomSeed;
 
     /**
      * Check the validation of input ion schema and execute the Ion Data generating process.
@@ -20,6 +22,11 @@ public class GeneratorOptions {
         String format = ((List<String>) optionsMap.get("--format")).get(0);
         String path = optionsMap.get("<output_file>").toString();
         String inputFilePath = optionsMap.get("--input-ion-schema").toString();
+        if (optionsMap.get("--seed") != null) {
+            randomSeed = new Random(Long.valueOf(optionsMap.get("--seed").toString()));
+        } else {
+            randomSeed = new Random();
+        }
         // Check whether the input schema file is valid and get the loaded schema.
         Schema schema = IonSchemaUtilities.loadSchemaDefinition(inputFilePath);
         ReadGeneralConstraints.constructAndWriteIonData(size, schema, format, path);
--- a/src/com/amazon/ion/benchmark/Main.java
+++ b/src/com/amazon/ion/benchmark/Main.java
@@ -252,7 +252,7 @@ public class Main {
             + "generated Ion data would conform with.\n"
 
         + "  -M --seed <long>      This option will be specified when users would like to get the same random data from the same schema file."
-            + "The provided value should be 64 bits of long seed value which will be used for creating pseudorandom number generator. \n"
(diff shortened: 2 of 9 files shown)


src/com/amazon/ion/benchmark/Main.java
```diff
@@ -251,6 +251,8 @@ public class Main {
         + "  -Q --input-ion-schema <file_path>      This option will specify the path of Ion Schema file which contains all constraints that the "
             + "generated Ion data would conform with.\n"
 
+        + "  -M --seed <seed_value>   This option will be specified when users would like to get the same random data from the same schema file.\n"
```
tgregg, line 254: We should be clear about what type of value is expected. An integer? What size?

src/com/amazon/ion/benchmark/ReadGeneralConstraints.java
```diff
@@ -14,6 +14,7 @@
 
 import java.io.FileOutputStream;
 import java.io.OutputStream;
+import java.util.Random;
```
tgregg, line 17: I don't see any other code changes in this file. Is this import necessary?

src/com/amazon/ion/benchmark/GeneratorOptions.java
```diff
@@ -4,11 +4,13 @@
 
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 
 /**
  * Execute Ion Data Generator after receiving the hashmap of command line options.
  */
 public class GeneratorOptions {
+    public static Random randomSeed;
```
tgregg, line 13: `randomSeed` doesn't feel like the right name for this. This is just a `Random` instance that may or may not have been seeded. `random` is probably a fine name.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -217,6 +261,24 @@ public void testCodepointLength() throws Exception {
...
+        assertTrue(generateAndCompare(true));
+    }
+
+    /**
+     * Test if the generated data is randomized when there is no '--seed' option provided.
+     * @throws Exception if error occurs during the executing and comparison process.
+     */
+    @Test
+    public void testWithoutSeed() throws Exception {
+        assertTrue(!generateAndCompare(false));
```
tgregg, line 279: Use `assertFalse` without the negation.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -126,6 +133,43 @@ public static void violationDetect(String inputFile) throws Exception {
...
+    public Boolean generateAndCompare(Boolean seedOption) throws Exception {
+        Boolean result = true;
+        for (File testFile : TEST_ISL_FILES) {
+            String testFilePath = testFile.getPath();
+            for (int i = 0; i < 2; i++) {
+                String outputFile = "testSeed" + i + ".ion";
+                Map <String, Object> optionsMap = Main.parseArguments("generate", "--data-size", "5000", "--format", "ion_text", "--input-ion-schema", testFilePath, outputFile);
+                if (seedOption == true) {
+                    optionsMap = Main.parseArguments("generate", "--data-size", "5000", "--seed", "200", "--format", "ion_text", "--input-ion-schema", testFilePath, outputFile);
+                }
```
tgregg, line 151: Consider using if/else to avoid redundantly calling `parseArguments` when `seedOption == true`.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
```diff
@@ -76,6 +80,9 @@ public class DataGeneratorTest {
     private final static String HEAP_USAGE = "Heap usage";
     private final static String SERIALIZED_SIZE = "Serialized size";
     private final static String SPEED = "speed";
+    private final static File[] TEST_ISL_FILES = new File("./tst/com/amazon/ion/datagenerator/").listFiles();
+    private final static String BEFORE = "testSeed0.ion";
+    private final static String AFTER = "testSeed1.ion";
```
tgregg, line 85: Looks like these constants aren't needed; they can be local to the `generateAndCompare` method.

tgregg (commented):
This is a great feature.

src/com/amazon/ion/benchmark/Main.java
linlin-s: > We should be clear about what type of value is expected. An integer? What size?

These information will be clarified in the following commit.

src/com/amazon/ion/benchmark/ReadGeneralConstraints.java
linlin-s: > I don't see any other code changes in this file. Is this import necessary?

Nice catch, will be deleted.

tst/com/amazon/ion/benchmark/DataGeneratorTest.java
linlin-s: > Consider using if/else to avoid redundantly calling `parseArguments` when `seedOption == true`.

Yes, that makes more sense. Will update it in the following commit.

src/com/amazon/ion/benchmark/Main.java
```diff
@@ -251,7 +251,8 @@ public class Main {
         + "  -Q --input-ion-schema <file_path>      This option will specify the path of Ion Schema file which contains all constraints that the "
             + "generated Ion data would conform with.\n"
 
-        + "  -M --seed <seed_value>   This option will be specified when users would like to get the same random data from the same schema file.\n"
+        + "  -M --seed <long>      This option will be specified when users would like to get the same random data from the same schema file."
+            + "The provided value should be 64 bits of long seed value which will be used for creating pseudorandom number generator. \n"
```
tgregg, line 255: "up to" 64 bits?

Issue #53, opened 2022-11-18
Account for a margin of error when comparing benchmark results


popematt:
As stable as the performance tests might be, they are not perfect. See for example, [this PR](https://github.com/amzn/ion-java/pull/463) that does not modify any of the library code, but failed the performance regression workflow because of a [detected regression](https://github.com/amzn/ion-java/actions/runs/3499274138/jobs/5860649031#step:22:32). We should determine some margin of error for the tests, and only flag a regression if it exceeds the margin of error.

I'm not sure the best way to make this change, but it is possible that change would need to be made here:
https://github.com/amzn/ion-java-benchmark-cli/blob/10807b6b3e06181752052da6f0febedf1b05f8ec/src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java#L262

Or here:
https://github.com/amzn/ion-java-benchmark-cli/blob/10807b6b3e06181752052da6f0febedf1b05f8ec/src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java#L179-L188

PR #54, opened 2022-12-05
Updates the method of calculating threshold value.

linlin-s:
Signed-off-by: Linlin Sun <linls@amazon.com>

*Issue #, if available:*
#53 
*Description of changes:*

This PR updates the methods of calculating the threshold of defining regression of ion-java performance. 
The original method of calculating threshold value is to find the largest relative difference among the data set of benchmark results from multiple iterations of benchmark operations. The threshold calculated by this strategy would cause false negatives. Even there’s no modification in ion-java library, the regression could still be detected sometimes. 

Before updating, the method of calculating threshold value used the maximum and the minimum values of the raw data set to calculate the relative difference which used as the threshold value of defining regression in the workflow.  In this PR, the method of calculating threshold applies [IQR](https://docs.oracle.com/en/cloud/saas/planning-budgeting-cloud/pfusu/insights_metrics_IQR.html)  method to find the lower fence and upper fence when calculating the threshold value.

In this PR, the time of benchmark iteration is updated to 20 times. This update will help to extend the size of data set which is used for calculating threshold value.

_Test:_
Since the method of calculating threshold value has been updated, the expected threshold values in the unit test should be updated too.

_Example:_
Performance regression detected test:
    This [commit](https://github.com/linlin-s/ion-java/pull/25/commits/5b6b29e7125e7c88500db74c6ea904fc9b0041f9) deletes the UTF8 String encoders to cause [writing performance regression ](https://github.com/linlin-s/ion-java/actions/runs/3634450152)of IonJava.
Performance regression not detected test:
    This [commit](https://github.com/linlin-s/ion-java/pull/25/commits/305ee067acd7f2df400b07001072fd246816d000) does not change the code in ion-java library and there is [no regression](https://github.com/linlin-s/ion-java/actions/runs/3634495854/jobs/6132627941) detected.


src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -99,8 +99,44 @@ private static BigDecimal getThresholdScore(IonList rawDataList) {
...
+     */
+    public static BigDecimal calculateThreshold(List<BigDecimal> sortedRawData) {
+        int length = sortedRawData.size();
+        BigDecimal Q1 = median(sortedRawData, 0, length / 2 - 1);
+        BigDecimal Q3 = median(sortedRawData, (length + 1) / 2, length - 1);
+        BigDecimal IQR = Q3.subtract(Q1);
+        BigDecimal range = IQR.multiply(BigDecimal.valueOf(1.5));
+        BigDecimal lowerFence = Q1.subtract(range);
+        BigDecimal upperFence = Q3.add(range);
+        return calculateDifference(upperFence, lowerFence);
```
tgregg, line 123: Is this what we want? Or do we want to compare the final result against the upper and lower fences to make sure it's between them? Does this somehow achieve that in a way that I'm not seeing?

tst/com/amazon/ion/benchmark/optionsCombinations.ion
```diff
@@ -1 +1 @@
-[" read --api dom --format ion_binary --results-format ion ", " read --api streaming --format ion_binary --results-format ion ", " write --api streaming --format ion_binary --results-format ion ", " read --api streaming --format json --results-format ion ", " read --api dom --format json --results-format ion ", " write --api streaming --format json --results-format ion "]
\ No newline at end of file
+[" read --api dom --format ion_binary --results-format ion --iterations 20 ", " read --api streaming --format ion_binary --results-format ion --iterations 20 ", " write --api streaming --format ion_binary --results-format ion --iterations 20 ", " read --api streaming --format json --results-format ion --iterations 20 ", " read --api dom --format json --results-format ion --iterations 20 ", " write --api streaming --format json --results-format ion --iterations 20 "]
```
tgregg, line 1: Is the increased number of iterations necessary for this technique to work? Do our Github Actions workflows run for that many iterations?

tgregg (commented):
This is an interesting technique. To ensure it's effective, we need to backtest it against previous results. You've ensured that it does not produce a false negative or false positive for one case each, but we should *also* get the data from previous false positive/negative results (we may only have instances of false positives - if so that's fine) and verify that this technique would not have produced the same inaccurate result. This should be possible in a unit testing environment - simply run the raw data points from a previous incorrect result through this new code and verify it correctly detects whether a regression occurred.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > Is this what we want? Or do we want to compare the final result against the upper and lower fences to make sure it's between them? Does this somehow achieve that in a way that I'm not seeing?

We are not trying compare the final result to lower and upper fences to see whether it falls into the range. The final result represent the relative difference between the benchmark results from two different commits. However, the lower and upper fences define the acceptable data fluctuate range of benchmark result from one commit.  In this way, lower and upper fences will help us to define the acceptable data difference among the raw data set. If the final results bigger than the relative difference calculated from lower fence and upper fence (which can be calculated by calling method `calculateDifference(upperFence, lowerFence)`), we will flag this comparison as regression detected since the final results already out of the acceptable relative difference.

PR #57, opened 2023-08-14
Fixes the issue that the '--version' option not return the expected version value of ion-java.

linlin-s:
### Issue #, if available:
N/A

### Description of changes:
The option '--version' of ion-java-benchmark-cli was not functional since the change of ion-java build. This PR fixed the issue to allow the option '--version' returns the correct ion-java version. 

----


tgregg (approved):
Thanks for fixing this!

PR #58, opened 2023-10-10
Update the performance regression detection method to reduce false positives.

linlin-s:
### Issue #, if available:
[#53](https://github.com/amazon-ion/ion-java-benchmark-cli/issues/53)
### Description of changes:

This PR optimized the method of detecting performance regression by making these changes:
1. Removing the logic of calculating threshold value of regression detection. Instead, using [two-sample t-test](https://en.wikipedia.org/wiki/Student%27s_t-test) to compare the benchmark results before and after changes to see if regression happened. More details about how to use t-test method to detect regression is included in the comments of source code.
2. Adding logic to remove outliers from raw data. Two-sample t-test requires normally distributed data. Given that some of our raw data deviates from this due to noise, we preprocess using the IQR ([Interquartile range](https://en.wikipedia.org/wiki/Interquartile_range)) method to filter out outliers. After this preprocessing, we verify data is normally distributed by using [Shapiro–Wilk test](https://en.wikipedia.org/wiki/Shapiro%E2%80%93Wilk_test).
3. Adding unit tests for methods `removeOutliers` and `detectRegression`.


src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
-            rawData.add(score.bigDecimalValue());
+    public static double detectRegression(double[] before, double[] after) {
+        TTest tTest = new TTest();
+        // Perform two-sample t-test
+        double pValue = tTest.tTest(before, after);
+        // Calculate means of both datasets
+        double meanBefore = StatUtils.mean(before);
+        double meanAfter = StatUtils.mean(after);
+        // Calculate the difference in means (regression value)
+        double regressionValue = (meanAfter - meanBefore) / meanBefore;
```
popematt, line 80: The code would be more logical/readable if these lines were moved into the `if (pValue <0.05) {` block.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
+     * @param before represents the benchmark results from benchmarking the previous ion-java.
+     * @param after represents the benchmark results from benchmarking the ion-java with the new changes.
+     * @return regression detection result. If the regression is detected, then the regression value will be returned, else 0 will be returned.
      */
-    private static BigDecimal getThresholdScore(IonList rawDataList) {
-        List<BigDecimal> rawData = new ArrayList<>();
-        for (int i = 0; i < rawDataList.size(); i++) {
-            IonDecimal score = (IonDecimal) rawDataList.get(i);
-            rawData.add(score.bigDecimalValue());
+    public static double detectRegression(double[] before, double[] after) {
```
popematt, line 72: Technically, this is detecting whether there is a statistically significant change. It looks like it will return a non-zero value both for a significant regression and a significant improvement.

src/com/amazon/ion/benchmark/.DS_Store
tgregg, line 1: This looks like an accidental commit?

src/com/amazon/ion/benchmark/Main.java
```diff
@@ -44,7 +44,7 @@ public class Main {
 
         + "  ion-java-benchmark run-suite (--test-ion-data <file_path>) (--benchmark-options-combinations <file_path>) <output_file>\n"
 
-        + "  ion-java-benchmark compare (--benchmark-result-previous <file_path>) (--benchmark-result-new <file_path>) <output_file>\n"
+        + "  ion-java-benchmark compare (--benchmark-result-previous <file_path>) (--benchmark-result-new <file_path>)\n"
```
tgregg, line 47: Was `<output_file>` not used? Or did you determine that it is not useful, and we should always write to `stdout`?

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
+        double upperBound = q3 + 1.5 * iqr;
+
+        List<Double> filteredData = new ArrayList<>();
+        for (double value : data) {
+            if (value >= lowerBound && value <= upperBound) {
+                filteredData.add(value);
+            }
         }
-        return scoreStruct;
+        return filteredData.stream().mapToDouble(d -> d).toArray();
```
popematt, line 135: We can make this a lot more concise. (I'm only 95% sure the syntax of suggestion is correct, but I know the concept is sound.)
```java
return Arrays.stream(data).filter(d -> lowerBound <= d && d <= upperBound).toArray();
```

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
+     * @return the processed data which is ready to be fed into two-sample T-test algorithm.
+     * @throws Exception if there's error occurred reading the benchmark result.
      */
-    public static BigDecimal getScore(String benchmarkResultFilePath, String keyWord) throws Exception {
-        IonStruct scoreStruct = parseScore(benchmarkResultFilePath, keyWord);
-        IonValue score = scoreStruct.get(SCORE);
-        if (score.getType().equals(IonType.FLOAT)) {
-            IonFloat scoreFloat = (IonFloat) score;
-            return scoreFloat.bigDecimalValue();
+    public static double[] preProcess(String benchmarkResult, String keyWord) throws Exception {
```
popematt, line 101: We should use a more descriptive name than `preProcess`. How about `loadKeywordSpecificBenchmarkResults`?

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
-     * @throws Exception if error occurs when reading Ion data.
+     * Converting IonList to DoubleStream.
+     * @param data represents the input IonList.
+     * @return the converted DoubleStream.
      */
-    private static IonValue getParameter(String benchmarkResultFilePath, String keyWord) throws Exception {
-        IonStruct benchmarkResultStruct = readHelper(benchmarkResultFilePath);
-        IonStruct parameterStruct = (IonStruct) benchmarkResultStruct.get(PARAMETERS);
-        return parameterStruct.get(keyWord);
+    public static DoubleStream toDouble(IonList data){
```
popematt, line 143: This can be private. You should also just return `List<Double>` or `double[]` unless you're going to be passing around `DoubleStream` to other functions. (E.g. if the input and output of `removeOutliers` was `DoubleStream`.)

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -1,51 +1,32 @@
...
     public static final List<String> BENCHMARK_SCORE_KEYWORDS = Arrays.asList("speed", "Heap usage", "Serialized size", "·gc.alloc.rate");
     private static final String PRIMARY_METRIC = "primaryMetric";
-    private static final String PARAMETERS = "params";
-    private static final String INPUT = "input";
-    private static final String OPTIONS = "options";
     private static final String SECONDARY_METRIC = "secondaryMetrics";
-    private static final String SCORE = "score";
     private static final String SPEED = "speed";
-    private final static String GC_ALLOCATE = "·gc.alloc.rate";
-    private final static String HEAP_USAGE = "Heap usage";
```
popematt, line 42: Can you explain to me why it's okay to get rid of these?

tgregg (approved):
Nice! The new technique is simpler (in the code we own, anyway) and more sound.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: Yes, sounds good to me. I'll update it in the next commit.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
```diff
@@ -57,110 +38,110 @@ public class ParseAndCompareBenchmarkResults {
...
+        Map<String, Double> comparisonResults = new HashMap<>();
         for (String benchmarkScoreKeyword : BENCHMARK_SCORE_KEYWORDS) {
-            BigDecimal previousScore = getScore(benchmarkResultPrevious, benchmarkScoreKeyword);
-            BigDecimal newScore = getScore(benchmarkResultNew, benchmarkScoreKeyword);
-            BigDecimal result = calculateDifference(previousScore, newScore);
-            scoreMap.put(benchmarkScoreKeyword, result);
+            double[] previousData = preProcess(benchmarkResultPrevious, benchmarkScoreKeyword);
+            double[] newData = preProcess(benchmarkResultNew, benchmarkScoreKeyword);
+            double comparisonResult = detectRegression(previousData, newData);
+            if (comparisonResult > 0) {
```
linlin-s, line 46: Replying to this [comment](https://github.com/amazon-ion/ion-java-benchmark-cli/pull/58/files/2b02d3a1cd79441eff09db1aa3a2ba4c23cc460e#r1355682971) 
`detectRegression()` method will return a non-zero value both for a significant regression and a significant improvement and only significant regression (when comparison result > 0) will be added to the final `comparisonResults` map.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > Technically, this is detecting whether there is a statistically significant change. It looks like it will return a non-zero value both for a significant regression and a significant improvement.

Yes, and only significant regression value will be added to the `comparisonResults` map which is used for deciding whether `ion-java-regression-detection` workflow should fail or not. Here is the [comment](https://github.com/amazon-ion/ion-java-benchmark-cli/pull/58/files/2b02d3a1cd79441eff09db1aa3a2ba4c23cc460e#r1355716936) about the conditional check

src/com/amazon/ion/benchmark/Main.java
linlin-s: Yes, `<output_file>` is not useful for now. We will write out the regression results.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: Thanks for the suggestion.

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: Thanks, will change to private in the next commit. 
The raw data is a nested `IonList`, and `toDouble` is used for transform the inner `IonList` to `DoubleStream` while flattening the raw data. The return value of `toDouble` is not directly passing to `removeOutliers`. [Here](https://github.com/amazon-ion/ion-java-benchmark-cli/pull/58/commits/2b02d3a1cd79441eff09db1aa3a2ba4c23cc460e#diff-53b4790715ee277df76aa3587d5ea213712796275b4394455ea8883b508b7aafR108-R110) is where the data flattening happened. In this case, should we still consider the returned data type as `List<Double>` or `double[]` ?

src/com/amazon/ion/benchmark/ParseAndCompareBenchmarkResults.java
linlin-s: > Can you explain to me why it's okay to get rid of these?

These two constant variables were used for composing the regression detection summary to represent which metric has regression. For now, we compose the regression detection summary by iterating the [comparisonResults](https://github.com/amazon-ion/ion-java-benchmark-cli/pull/58/commits/2b02d3a1cd79441eff09db1aa3a2ba4c23cc460e#diff-53b4790715ee277df76aa3587d5ea213712796275b4394455ea8883b508b7aafR41). The map `comparisonResults` contains key-value pairs where the keys are metrics that have regression detected and the values are the amounts they have regressed. The keys already include the names of deleted variables, so the deleted information does not need to be initialized separately.

popematt (approved):

PR #60, opened 2024-05-08
Adds support for binary Ion 1.1.

tgregg:
### Issue #, if available:

https://github.com/amazon-ion/ion-java/issues/747

### Description of changes:

* Adds the `--ion-minor-version` option.
* Supports conversion between Ion minor versions.
* Tests binary Ion 1.1.

Note: I propose merging this to branch `ion-1-1`, because the change in this PR is only buildable with the `ion-11-encoding` branch of ion-java for now.

----


src/com/amazon/ion/benchmark/OptionsCombinationBase.java
```diff
@@ -51,6 +52,7 @@ abstract class OptionsCombinationBase {
     final int limit;
     final boolean jsonUseBigDecimals;
     final boolean autoFlush;
+    final Integer ionMinorVersion;
```
popematt, line 55: I wish we could just use `ionVersion` and accept a string such as "1.1" or "1.0", but I get that this is much simpler.

popematt (approved):

tst/com/amazon/ion/benchmark/OptionsTest.java
```diff
@@ -2102,4 +2107,128 @@ public void testDetectRegression() {
...
+        assertReadTaskExecutesCorrectly(
+            "binaryAllTypes11.10n",
+            optionsCombination,
+            Format.ION_BINARY,
+            false
+        );
+    }
+
+    @Test
+    public void readIon11WithConversionFrom10() throws Exception {
```
linlin-s, line 2130: I'm wondering if the condition would happen when the input data is 1.1 encoded but the minor version number is specified as 0.

tgregg: Yes, it will first convert the 1.1 data to 1.0 and then read. I've added testing of this case in revision 2 of the PR.

PR #64, opened 2024-06-05
Allows symbol table boundaries to be preserved when converting from Ion 1.0 to Ion 1.1.

### LINKED ISSUE #62: Ensure Ion 1.1 system values are transcribed verbatim
E.g. symbol tables should occur in the same locations, with the same number of user values in their scopes. This will provide a more fair comparison to Ion 1.0, which does this.

tgregg:
### Issue #, if available:
Fixes #62 

### Description of changes:
This allows more direct performance comparisons between Ion 1.0 and Ion 1.1. Note: ion-java's 1.1 writer needs to flush upon encountering a system value in order to achieve comparable performance to the Ion 1.0 writer (see https://github.com/amazon-ion/ion-java/issues/873).

----

### DIFF
--- a/src/com/amazon/ion/benchmark/IonUtilities.java
+++ b/src/com/amazon/ion/benchmark/IonUtilities.java
@@ -11,6 +11,7 @@ import com.amazon.ion.SpanProvider;
 import com.amazon.ion.SymbolTable;
 import com.amazon.ion.impl._Private_IonConstants;
 import com.amazon.ion.impl._Private_IonSystem;
+import com.amazon.ion.impl._Private_IonWriter;
 import com.amazon.ion.impl.bin.DelimitedContainerStrategy;
 import com.amazon.ion.impl.bin.SymbolInliningStrategy;
 import com.amazon.ion.impl.bin._Private_IonManagedBinaryWriterBuilder;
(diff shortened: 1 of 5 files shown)


src/com/amazon/ion/benchmark/IonUtilities.java
```diff
@@ -419,8 +447,9 @@ static void rewriteIonFile(Path input, Path output, OptionsCombinationBase optio
                 options.importsForInputFile == null &&
                 options.importsForBenchmarkFile == null &&
                 options.format == Format.ION_BINARY &&
-                options.ionMinorVersion != 1 && // TODO remove once support for writing system values via the Ion 1.1 writer is added.
-                IonUtilities.minorVersionsEqual(options.ionMinorVersion, input.toFile())
+                // Minor versions may add new kinds of system values, so it is not possible to maintain system value
```
popematt, line 483: Minor versions could also add new symbols to the system symbol table. Do we need to account for that somehow to make sure that the transcoded SIDs still point to the correct text?

tgregg: Oof, yeah... looking into whether this is feasible.

tgregg: Proposed update to ion-java to enable this here: https://github.com/amazon-ion/ion-java/pull/876
And I updated this PR to perform the transcoding. We'll have to make a minor update to ion-java once the 1.1 system symbol table is finalized.

popematt (approved):

PR #69, opened 2025-07-28
Add Support for IonElement

zihaowang678:
**Summary**
This PR adds support for the `ION_ELEMENT_DOM` API to the ion-java-benchmark-cli, enabling benchmarking of the IonElement library from ion-element-kotlin alongside the existing STREAMING and DOM APIs.

**Changes**
1. API Enum Enhancement (`API.java`)
- Added `ION_ELEMENT_DOM` enum value with documentation

2. Read Task Implementation (`IonMeasurableReadTask.java`)
- Implemented element reading methods for buffer and file operations
- Added proper resource management and side effect consumer integration

3. Write Task Implementation (`IonMeasurableWriteTask.java`)
- Added `generateWriteInstructionsElement()` method
- Integrated element processing with limits and flush period support

4. Task Routing (`MeasurableReadTask.java` & `MeasurableWriteTask.java`)
- Added `ION_ELEMENT_DOM` routing in `getTask()` and `setUpTrial()` methods

5. CLI Documentation (`Main.java`)
- Updated `--api` option to include `ion_element_dom` with format restrictions

6. Dependencies (`pom.xml`)
- Added ion-element (1.3.0) and Kotlin dependencies

7. Test Coverage (`OptionsTest.java`)
- Added comprehensive tests for read/write operations with various configurations

**Performance Analysis**
Evaluated on three Ion corpus datasets, including both Ion text and Ion binary data, IonElement shows an average of ~37.09% increase in `primary_score` of performance compared to IonValue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,17 +101,28 @@ ion-java-benchmark write --ion-imports-for-benchmark tables.ion \
 Benchmark a full-traversal read of data equivalent to exampleWithImports.10n, which declares the shared
 symbol table imports provided by inputTables.ion, re-encoded (if necessary) using the shared symbol
 tables provided by benchmarkTables.ion, inputTables.ion, and no shared symbol tables. Produce
-results from using both the DOM and IonReader APIs.
+results from using the DOM, IonReader and IonElement APIs.
 
 ```
 ion-java-benchmark read --ion-imports-for-input inputTables.ion \
                         --ion-imports-for-benchmark benchmarkTables.ion \
                         --ion-imports-for-benchmark auto \
                         --ion-imports-for-benchmark none \
-                        --ion-api dom \
-                        --ion-api streaming \
+                        --api dom \
+                        --api streaming \
+                        --api ion_element_dom \
                         exampleWithImports.10n
 ```
+
+Benchmark a full-traversal read of `example.10n` using the IonElement API from ion-element-kotlin,
+comparing performance against the traditional DOM API.
+
+```
+ion-java-benchmark read --api dom \
+                        --api ion_element_dom \
+                        example.10n
+```
+
 
 ## Tips
 
--- a/pom.xml
+++ b/pom.xml
@@ -100,7 +100,7 @@
         <dependency>
             <groupId>com.amazon.ion</groupId>
             <artifactId>ion-element</artifactId>
-            <version>1.3.0</version>
+            <version>[1.3.0, )</version>
         </dependency>
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,21 @@
             <artifactId>ion-java-path-extraction</artifactId>
             <version>1.2.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.amazon.ion</groupId>
+            <artifactId>ion-element</artifactId>
+            <version>1.3.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
+            <version>1.6.20</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlinx</groupId>
+            <artifactId>kotlinx-collections-immutable-jvm</artifactId>
+            <version>0.3.4</version>
+        </dependency>
         <!-- test dependencies -->
         <dependency>
           <groupId>junit</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -132,6 +132,13 @@
                 <configuration>
                     <source>8</source>
                     <target>8</target>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
+                            <version>1.23</version>
+                        </path>
+                    </annotationProcessorPaths>
                 </configuration>
             </plugin>
             <plugin>
--- a/src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
+++ b/src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
@@ -150,10 +150,11 @@ public class CborJacksonMeasurableReadTask extends MeasurableReadTask {
 
     @Override
     void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException {
-        sideEffectConsumer = consumer;
-        CBORMapper objectMapper = new CBORMapper(cborFactory);
-        JsonNode tree = objectMapper.readTree(options.newInputStream(inputFile));
-        consumer.consume(tree);
+        CBORMapper mapper = JacksonUtilities.newCborObjectMapper(cborFactory, options);
+        Iterator<JsonNode> iterator = mapper.reader().createParser(options.newInputStream(inputFile)).readValuesAs(JsonNode.class);
+        while (iterator.hasNext()) {
+            consumer.consume(iterator.next());
+        }
     }
 
     @Override
--- a/src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
+++ b/src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
@@ -150,10 +150,19 @@ public class CborJacksonMeasurableReadTask extends MeasurableReadTask {
 
     @Override
     void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException {
-        CBORMapper mapper = JacksonUtilities.newCborObjectMapper(cborFactory, options);
-        Iterator<JsonNode> iterator = mapper.reader().createParser(options.newInputStream(inputFile)).readValuesAs(JsonNode.class);
-        while (iterator.hasNext()) {
-            consumer.consume(iterator.next());
-        }
+        sideEffectConsumer = consumer;
+        CBORMapper objectMapper = new CBORMapper(cborFactory);
+        JsonNode tree = objectMapper.readTree(options.newInputStream(inputFile));
+        consumer.consume(tree);
+    }
+
+    @Override
+    public void fullyReadElementFromBuffer(SideEffectConsumer consumer) throws IOException {
+        throw new UnsupportedOperationException("IonElement API is not supported for CBOR format. Use ion_binary or ion_text format instead.");
+    }
+
+    @Override
+    public void fullyReadElementFromFile(SideEffectConsumer consumer) throws IOException {
+        throw new UnsupportedOperationException("IonElement API is not supported for CBOR format. Use ion_binary or ion_text format instead.");
     }
 }
--- a/src/com/amazon/ion/benchmark/IonMeasurableReadTask.java
+++ b/src/com/amazon/ion/benchmark/IonMeasurableReadTask.java
@@ -9,7 +9,7 @@ import com.amazon.ionpathextraction.PathExtractor;
 import com.amazon.ionpathextraction.PathExtractorBuilder;
 import com.amazon.ionelement.api.AnyElement;
 import com.amazon.ionelement.api.IonElementLoader;
-import com.amazon.ionelement.api.IonElementLoaderOptions;
+import com.amazon.ionelement.api.ElementLoader;
 
 import java.io.IOException;
 import java.nio.file.Path;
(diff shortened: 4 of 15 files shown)


README.md
```diff
@@ -101,18 +101,29 @@ ion-java-benchmark write --ion-imports-for-benchmark tables.ion \
...
 ```
 
+Benchmark a full-traversal read of `example.10n` using the IonElement API from ion-element-kotlin,
+comparing performance against the traditional DOM API.
+
+```
+ion-java-benchmark read --api dom \
+                        --api ion_element_dom \
+                        example.10n
+```
```
popematt, line 124: Just curious—when you run this, what sort of results to you get?

src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
```diff
@@ -150,10 +150,19 @@ void fullyReadDomFromBuffer(SideEffectConsumer consumer) throws IOException {
...
     void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException {
-        CBORMapper mapper = JacksonUtilities.newCborObjectMapper(cborFactory, options);
-        Iterator<JsonNode> iterator = mapper.reader().createParser(options.newInputStream(inputFile)).readValuesAs(JsonNode.class);
-        while (iterator.hasNext()) {
-            consumer.consume(iterator.next());
-        }
+        sideEffectConsumer = consumer;
+        CBORMapper objectMapper = new CBORMapper(cborFactory);
+        JsonNode tree = objectMapper.readTree(options.newInputStream(inputFile));
+        consumer.consume(tree);
```
popematt: The reason for changing this method is not obvious to me since it seems unrelated to IonElement. Can you explain the purpose of this change?

src/com/amazon/ion/benchmark/IonMeasurableReadTask.java
```diff
@@ -67,6 +71,13 @@ private int pathExtractorCallback(IonReader reader) {
         } else {
             reusableLobBuffer = null;
         }
+
+        // Initialize IonElement loader for ION_ELEMENT_DOM API
+        if (options.api == API.ION_ELEMENT_DOM) {
+            elementLoader = new com.amazon.ionelement.impl.IonElementLoaderImpl(
+                    new IonElementLoaderOptions(false)
+            );
```
popematt: Instead of using these constructors (which shouldn't be visible—we'll have to look into that), you should use [the factory method that is part of the public API](https://github.com/amazon-ion/ion-element-kotlin/blob/5bd70e045ba31a9ae3096ad44e24a1698592f721/src/main/kotlin/com/amazon/ionelement/api/IonElementLoader.kt#L221).

```suggestion
            elementLoader = ElementLoader.createIonElementLoader();
```
Also note that the default is to _not_ include the location metadata, so you don't have to explicitly pass in that option.

src/com/amazon/ion/benchmark/IonMeasurableWriteTask.java
```diff
@@ -39,6 +45,13 @@ class IonMeasurableWriteTask extends MeasurableWriteTask<IonWriter> {
         } else {
             throw new IllegalStateException("IonFormatWriter is compatible only with ION_TEXT and ION_BINARY");
         }
+
+        // Initialize IonElement loader for ION_ELEMENT_DOM API
+        if (options.api == API.ION_ELEMENT_DOM) {
+            elementLoader = new com.amazon.ionelement.impl.IonElementLoaderImpl(
+                    new IonElementLoaderOptions(false)
+            );
```
popematt:
```suggestion
            elementLoader = ElementLoader.createIonElementLoader();
```

src/com/amazon/ion/benchmark/IonMeasurableWriteTask.java
```diff
@@ -195,4 +208,35 @@ public void closeWriter(IonWriter writer) throws IOException {
...
+            }
+        } else {
+            List<AnyElement> limitedElements = new ArrayList<>();
+            try (IonReader reader = IonUtilities.newReaderBuilderForInput(options).build(options.newInputStream(inputFile))) {
+                Iterable<AnyElement> allElements = elementLoader.loadAllElements(reader);
+                int count = 0;
+                for (AnyElement element : allElements) {
+                    limitedElements.add(element);
+                    if (++count >= options.limit) break;
+                }
```
popematt: So, what you're doing here is loading all elements, and then copying them one-by-one into a shorter list. One of the benefits of setting a limit is that if the data stream is too large, you can still load a subset of the data stream. However, this implementation will break if there is a data stream that is too large for the backing implementation. Instead, you can do something like this to ensure that we don't load more than the requested number of items from the data stream.

```suggestion
                int count = 0;
                while (count++ < options.limit && reader.next() != null) {
                    limitedElements.add(elementLoader.loadCurrentElement(reader));
                }
```

src/com/amazon/ion/benchmark/JsonJacksonMeasurableReadTask.java
```diff
@@ -139,10 +139,19 @@ void fullyReadDomFromBuffer(SideEffectConsumer consumer) throws IOException {
...
     void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException {
-        ObjectMapper mapper = JacksonUtilities.newJsonObjectMapper(jsonFactory, options);
-        Iterator<JsonNode> iterator = mapper.reader().createParser(options.newInputStream(inputFile)).readValuesAs(JsonNode.class);
-        while (iterator.hasNext()) {
-            consumer.consume(iterator.next());
-        }
+        sideEffectConsumer = consumer;
+        ObjectMapper objectMapper = JacksonUtilities.newJsonObjectMapper(jsonFactory, options);
+        JsonNode tree = objectMapper.readTree(options.newInputStream(inputFile));
+        consumer.consume(tree);
```
popematt: Can you explain the purpose of changing this method?

src/com/amazon/ion/benchmark/MeasurableReadTask.java
```diff
@@ -84,6 +84,20 @@ abstract class MeasurableReadTask implements MeasurableTask {
      */
     abstract void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException;
 
+    /**
+     * Initialize the element loader and perform a fully-materialized deep read of the data from a buffer using
+     * IonElement API. The "loader" is defined as any context that is tied to a single stream.
+     * @throws IOException if thrown during reading.
+     */
+    abstract void fullyReadElementFromBuffer(SideEffectConsumer consumer) throws IOException;
```
popematt, line 92: Not a blocker—it would be nice if we could add new APIs to the benchmark CLI without having to add more methods for each one. (Especially since they are unlikely to be applicable for all data formats.) Have you thought about any ways this could be factored to make it more easily extensible?

pom.xml
```diff
@@ -97,6 +97,21 @@
             <artifactId>ion-java-path-extraction</artifactId>
             <version>1.2.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.amazon.ion</groupId>
+            <artifactId>ion-element</artifactId>
+            <version>1.3.0</version>
```
popematt: As with `ion-java`, let's take the latest available version.
```suggestion
            <version>[1.3.0, )</version>
```

popematt (commented):
I have a few questions and suggestions, but overall this is looking pretty good.

README.md
zihaowang678: It runs two APIs in sequence, then gives two stacked statistical summaries at the end, one for each API.

src/com/amazon/ion/benchmark/CborJacksonMeasurableReadTask.java
zihaowang678: That was an accidental change from when I was debugging the IonElement integration. I'll revert it since it's unrelated.

src/com/amazon/ion/benchmark/IonMeasurableWriteTask.java
zihaowang678: Thanks for catching this.

src/com/amazon/ion/benchmark/JsonJacksonMeasurableReadTask.java
zihaowang678: Same as the CBOR change. I'll revert it since it's unrelated.

src/com/amazon/ion/benchmark/MeasurableReadTask.java
zihaowang678: You're right, the current approach is a straightforward one but doesn't scale well. A capability-based system where each format declares which APIs it supports might be better. I'd be happy to refactor this in a follow-up if it's worth doing.

src/com/amazon/ion/benchmark/IonMeasurableReadTask.java
```diff
@@ -251,4 +260,26 @@ public void fullyReadDomFromFile(SideEffectConsumer consumer) throws IOException
...
     }
+
+    @Override
+    public void fullyReadElementFromBuffer(SideEffectConsumer consumer) throws IOException {
+        sideEffectConsumer = consumer;
+        IonReader reader = readerBuilder.build(buffer);
+        Iterable<AnyElement> elements = elementLoader.loadAllElements(reader);
+        for (AnyElement element : elements) {
+            consumer.consume(element);
+        }
```
tgregg, line 275: It occurs to me that we should add a similar loop to `fullyReadDomFromFile` and `fullyReadDomFromBuffer`, both to protect against incorrect optimizations and to provide an apples-to-apples comparison between IonValue and IonElement.

src/com/amazon/ion/benchmark/Main.java
```diff
@@ -133,10 +133,10 @@ public class Main {
         + "  -f --format <type>                     Format to benchmark, from the set (ion_binary | ion_text | json | "
             + "cbor). May be specified multiple times to compare different formats. [default: ion_binary]\n"
 
-        + "  -a --api <api>                         The API to exercise (dom or streaming). For the ion-binary or "
+        + "  -a --api <api>                         The API to exercise (dom, streaming, or ion_element_dom). For the ion-binary or "
             + "ion-text formats, 'streaming' causes IonReader/IonWriter to be used while 'dom' causes IonLoader to be "
-            + "used. For Jackson JSON, 'streaming' causes JsonParser/JsonGenerator to be used while 'dom' causes "
-            + "ObjectMapper to materialize JsonNode instances. May be specified multiple times to compare both "
+            + "used, and 'ion_element_dom' causes IonElement from ion-element to be used with dedicated code paths. For Jackson JSON, 'streaming' causes JsonParser/JsonGenerator to be used while 'dom' causes "
```
tgregg:
```suggestion
            + "used, and 'ion_element_dom' causes IonElement from ion-element-kotlin to be used with dedicated code paths. For Jackson JSON, 'streaming' causes JsonParser/JsonGenerator to be used while 'dom' causes "
```

README.md
popematt: Sorry, I should have been more clear. Approximately how much of a performance difference are you getting between the two APIs when you run this command?

tst/com/amazon/ion/benchmark/OptionsTest.java
```diff
@@ -718,6 +758,51 @@ public void readBothTextAndIonUsingBothDomAndReader() throws Exception {
         assertTrue(expectedCombinations.isEmpty());
     }
 
+    @Test
+    public void readBothTextAndIonUsingIonElementDom() throws Exception {
```
tgregg:
```suggestion
    public void readBothTextAndBinaryUsingIonElementDom() throws Exception {
```
Note: I see that I initially created this naming pattern with the method above. Oops!

tst/com/amazon/ion/benchmark/OptionsTest.java
```diff
@@ -782,6 +867,37 @@ public void readBinaryWithLimitFromFileUsingDom() throws Exception {
         assertTrue(expectedCombinations.isEmpty());
     }
 
+    @Test
+    public void readBinaryWithLimitFromFileUsingIonElementDom() throws Exception {
```
tgregg:
```suggestion
    public void readBinaryAndTextWithLimitFromFileUsingIonElementDom() throws Exception {
```

tgregg (commented):
It's looking good! Note the failing CI build.

README.md
zihaowang678: Approximately a 30-ish percentage improvement in `primary_score`. While this command is a rough comparison between two APIs, the result aligns well with our more extensive experiments run in `SampleTime` mode across three ion datasets, where we saw a ~37.09% improvement in `primary_score`.

src/com/amazon/ion/benchmark/IonMeasurableReadTask.java
zihaowang678: Thanks for pointing this out! Will do.

tgregg (approved):
I'm excited to be able to use this to make the case for adoption of ion-element-kotlin over IonValue.

Once @popematt approves, we can squash/merge.

popematt (approved):

PR #70, opened 2025-10-07
Compare normalized allocation rate instead of non-normalized in compare command

austnwil:
### Description of changes:

`ion-java-benchmark-cli compare ...` will report regressions on average memory allocation rate/second (gc.alloc.rate), which is not a signal worth relying on since it is sensitive to execution speed. This changes the compare command to compare normalized allocation rate instead and adds a quick test.

This should reduce some false positives in the ion-java performance regression GitHub workflow, since I've noticed many of them are related to gc alloc rate.


tgregg (approved):
Good find

jobarr-amzn (approved):
Thanks!

